PR #5, opened 2026-03-20
Hookify Enhancements

Caladius:
Adds Hooks and CVars to the scattered #ifdef ENHANCEMENT marks around the code.


src/core1/viewport.c
```diff
@@ -344,13 +344,11 @@ bool viewport_cube_isInFrustum2(Cube *cube) {
     rel_pos[2] = (f32) ((cube->z * 1000) + 500) - sViewportPosition[2];
 
     // [port] Draw distance cutoff — skip when extended draw distance is enabled
-#ifdef ENHANCEMENT
-    // TODO: replace with CVar-driven multiplier
-#else
-    if (LENGTH_SQ_VEC3F(rel_pos) > 1.6e7f) {
-        return false;
+    if (!CVarGetInteger(CVAR_ENHANCEMENT("Graphics.DrawDistance"), 0)) {
```
JeodC, line 347: We'll have to add more to this for widescreen support with no enhancement, but for the enhancement side this might be ok. Needs proper testing because some models and animations will have odd behavior when still visible.

src/core2/cutscene/lair.c
```diff
@@ -29,16 +29,14 @@ u8 D_80383190;
 /* .code */
 // func_8031C640
 bool cutscene_skipIntroCutsceneCheck(void) {
-#ifdef ENHANCEMENT
-    // [port] Always allow Start to skip intro cutscenes
-    if (func_8024E698(0) == 1) {
+    // [port] Skip intro cutscene
+    if (CVarGetInteger(CVAR_ENHANCEMENT("Cutscenes.StartSkipIntro"), 0) && func_8024E698(0) == 1) {
```
JeodC, line 33: This should be two options or the enhancement allows pressing Start. If it's enabled the user should have the *option* to skip it. The `if (func_8024E698(0) == 1) {` check is what checks the Start button. You might also want to add this for the other unskippable scenes in this file including the first time the player views the Game Over screen. The game does not actually flush save to disk until the Game Over scene ends and reloads the Rareware Intro, so it will probably be desired.

src/core2/gc/pauseMenu.c
```diff
@@ -762,13 +762,9 @@ s32 gcpausemenu_initLargestPageIndex(void) {
...
 }
 
-bool gcpausemenu_initReturnToLair(void) {
-#ifdef ENHANCEMENT
-    s32 level = level_get();
-    return !(level > 0 && level < LEVEL_C_BOSS && D_8036C560[level - 1].map != -1);
-#else
-    return true;
-#endif
+bool gcpausemenu_initReturnToLair(bool shouldInit) {
```
JeodC: This needs testing that Return to Lair may just return the player to 1F lair entrance instead of mimicking world exit. That might be VB if this feature had actually been implemented though...since it's becoming an enhancement perhaps we should just go ahead and make it mimic world exit.

JeodC: #8 swaps the current ETL implementation for ProxySaw's simplified version. Sorry for the hassle. I did test it and it does return the player to the world entrance.

src/port/ui/LighthouseMenu.h
```diff
@@ -21,6 +21,7 @@ class LighthouseMenu : public Ship::Menu {
     WidgetInfo& AddWidget(WidgetPath& pathInfo, std::string widgetName, WidgetType widgetType);
     void AddMenuSettings();
     void AddMenuEnhancements();
+    void AddMenuRestorations();
```
Malkierian: Why is this a thing? Aren't they supposed to be in enhancements?

Caladius: This is restoring unused code that already existed. Specifically Return To Lair menu option

Malkierian: Right, but SoH and 2ship have Restorations entries in Enhancements.

Caladius: Why did I think it was separate?

Ugh

Malkierian (approved):

PR #7, opened 2026-03-20
Add rumble support

JeodC:


src/core1/ba_motor.c
```diff
@@ -107,32 +107,44 @@ void __baMotor_80250BA4(s32 arg0, s32 arg1, s32 arg2){
...
-    func_8024F35C(0);
-    D_802823AC = (pfs_status == 0);
-    D_802823B0 = D_802823AC;
-    if(D_802823AC){
-        osCreateMesgQueue(&D_80282390, &D_802823A8, 1);
-        osCreateThread(&sRumbleThread, 8, rumbleThread_entry, NULL, sRumbleThreadStack + RUMBLE_THREAD_STACK_SIZE, 25);
-        osStartThread(&sRumbleThread);
-        viMgr_func_8024BDAC(&D_80282390, OS_MESG_32(NULL));
-    }
-#endif
```
Malkierian, line 135: I'd prefer not completely removing decomp code like this. Leaving the ifdef should be fine, it'll be cleaner than leaving it in a comment. Don't want to lose the decomp footprint until we have full documentation.

src/core2/cutscene/lair.c
```diff
@@ -29,7 +29,7 @@ u8 D_80383190;
 /* .code */
 // func_8031C640
 bool cutscene_skipIntroCutsceneCheck(void) {
-#ifdef ENHANCEMENT
+#ifndef ENHANCEMENT
```
Malkierian: I think this wasn't meant to come in here? It would cause conflicts with Cal's PR too.

JeodC: No it was not, after recreating my fork I had to undef it to progress in a new save file (my fault) and forgot to swap it back.

src/core1/ba_motor.c
JeodC: Restored.

Malkierian (approved):

PR #21, opened 2026-03-22
Replaced ton of missing symbols

KiritoDv:


include/functions.h
```diff
@@ -32,20 +32,20 @@
...
 #pragma intrinsic (fabsf)
 
-//known it uses "+" instead of "|" for fight/code_9D40.c, func_8039049C, case 6
+//known it uses "+" instead of "|" for fight/code_9D40.c, chjinjonator_update, case 6
 #define FUNC_8030E624(sfx_e, vol, sample_rate) func_8030E624(\
     _SHIFTL((vol*1023), 21, 11) + _SHIFTL(sample_rate >> 5, 11, 10) + _SHIFTL(sfx_e, 0, 11)\
 )
 
-#define FUNC_8030E8B4(sfx_e, vol, sample_rate, position, e, f) func_8030E8B4(\
+#define FUNC_8030E8B4(sfx_e, vol, sample_rate, position, e, f) sfx_playFadeShorthand(\
```
Malkierian: Should this define and references to it also be changed to match the function name change?

Malkierian: Looks like the only one like it in this changeset.

include/functions.h
```diff
@@ -32,20 +32,20 @@
...
 #pragma intrinsic (fabsf)
 
-//known it uses "+" instead of "|" for fight/code_9D40.c, func_8039049C, case 6
+//known it uses "+" instead of "|" for fight/code_9D40.c, chjinjonator_update, case 6
 #define FUNC_8030E624(sfx_e, vol, sample_rate) func_8030E624(\
     _SHIFTL((vol*1023), 21, 11) + _SHIFTL(sample_rate >> 5, 11, 10) + _SHIFTL(sfx_e, 0, 11)\
 )
 
-#define FUNC_8030E8B4(sfx_e, vol, sample_rate, position, e, f) func_8030E8B4(\
+#define SFX_PlayFade(sfx_e, vol, sample_rate, position, e, f) SFX_PlayFade(\
```
Malkierian: I don't think the define is supposed to be the same casing as the function.

src/core1/audio/n_synthesizer.c
```diff
@@ -24,7 +24,7 @@ extern u32 client_num, client_cnt, client_max, client_min;
 #   define MIN(a,b) (((a)<(b))?(a):(b))
 #endif
 
-static s32 func_8025C370(ALPlayer **client);
+static s32 _n_timeToSamplesNoRound(ALPlayer **client);
 static s32 _n_timeToSamplesNoRound(s32 micros);
```
Malkierian: This got double defined with that rename.

src/core2/collision/climbsurface.c
```diff
@@ -5,10 +5,10 @@
 #include "core2/ba/physics.h"
 
 
-extern int        func_80258424(f32 vec[3], f32 minX, f32 minY, f32 minZ, f32 maxX, f32 maxY, f32 maxZ);
+extern int        ml_vec3f_inside_box_f(f32 vec[3], f32 minX, f32 minY, f32 minZ, f32 maxX, f32 maxY, f32 maxZ);
```
Malkierian: This getting changed now conflicts with a function of the same name in `ml.c`, but which is bool instead of int. Error comes back "redifinition; different basic types".

KiritoDv: Fixed

src/fight/chfinalboss.c
```diff
@@ -8,9 +8,9 @@
 extern void func_8028F4B8(f32[3], f32, f32);
 extern void func_80320ED8(ActorMarker *, f32, s32);
 s32 func_803297C8(Actor*, f32*);
-Actor *func_8032813C();
-extern Actor* func_80329958(ActorMarker *this, s32 arg1);
-extern void func_803298D8();
+Actor *actor_spawnWithYaw_f32();
+extern Actor* marker_getActor(ActorMarker *this, s32 arg1);
```
Malkierian: This conflicts with the declaration in `functions.h`. Different argument counts. Assuming they're supposed to be different functions.

src/MMM/minigame_organinput.c
```diff
@@ -214,7 +214,7 @@ void func_8038A964(void){
     }
 }
 
-void func_8038A994() {
+void chvilegame_get_grumblie_model() {
```
Malkierian: Different base type with extern declaration in `ch/yumblie.c`.

JeodC: And Mr Vile's game is in BGS. So are the filenames incorrect/misleading? On closer inspection this appears to be a file contextualized with code for Tumbar's Shed.

PR #40, opened 2026-04-02
Centralize port patches and add event hooks

JeodC:
- Create `Patches.h` as central header for all `port_*` function declarations, replacing scattered extern declarations across 16 decomp files
- Move framebuffer state, frame pacing, and sprite display cache out of `Game.cpp` and decomp files into dedicated patch .cpp files
- Add `OnMapLoad`, `OnFileLoaded` events and `VB_STATIC_CAMERA_SET/EXIT` behaviors so camera, save, and window title logic use the event system instead of direct port_* calls from decomp
- Split `port_shouldCaptureTransition()` out of transition.c into accessor functions in gctransition.h
- Preserve original decomp code in #if 0 blocks where it was mistakenly replaced


src/core1/io/pfs_manager.c
```diff
@@ -246,34 +240,13 @@ void pfsManager_update(void) {
...
-            sDemoViCount = displayViCount;
-        }
+        // [port] Override display pacing for maps that ran slow on N64.
+        port_setDemoViCount(port_getDemoDisplayViCount(sp5C));
     } else {
         // [port] Use the VI divisor from cutscene framerate actors so Game.cpp
         // paces the display correctly for slower cutscenes.
         s32 viDiv = viMgr_func_8024BFA0();
-        sDemoViCount = (viDiv > 2) ? viDiv : 0;
+        port_setDemoViCount((viDiv > 2) ? viDiv : 0);
```
Malkierian, line 249: Why does this function even exist, let alone get used here?

src/core2/anim/anim_bonetransform.c
```diff
@@ -361,57 +361,27 @@ bool asset_isCompressed(enum asset_e arg0){ //asset_compressed?
...
     BKSprite *s0;
     s0 = assetcache_get(sprite_id);
+#if 0 // [port] original decomp — assetCacheCurrentIndex is never updated in the port
+    if(D_80383CD4[assetCacheCurrentIndex] == NULL){
+        codeAEDA0_setSpriteDrawMode(-1);
+        func_80338308(sprite_getUnk8(s0), sprite_getUnkA(s0));
+        D_80383CD4[assetCacheCurrentIndex] = func_80344A1C(s0);
+    }
+    *arg1 = D_80383CD4[assetCacheCurrentIndex];
+#else
```
Malkierian: Shouldn't this be an endif, since the code after this should run whether this block gets run or not?

src/port/patches/CameraPatches.cpp
```diff
@@ -0,0 +1,114 @@
...
+            rotation[1] += sWsYawFixes[i].adjust;
+            break;
+        }
+    }
+}
+
+// Event listeners
+
+void RegisterCameraPatches_Init() {
+    REGISTER_LISTENER(OnMapLoad, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
```
Malkierian, line 103: Because these are in ShipInit, they should be using COND_HOOK (in this instance. The other one below should be using COND_VB_SHOULD), as they automatically set up a listener de-registration to prevent duplicate listeners.

src/port/patches/CameraPatches.cpp
```diff
@@ -0,0 +1,114 @@
...
+        if (ev->id == VB_STATIC_CAMERA_SET) {
+            int32_t nodeIndex = *(int32_t*)ev->args;
+            sLastStaticCameraNode = nodeIndex;
+        } else if (ev->id == VB_STATIC_CAMERA_EXIT) {
+            sLastStaticCameraNode = -1;
+        }
+    });
+}
+
+static RegisterShipInitFunc initFunc(RegisterCameraPatches_Init);
```
Malkierian, line 123: This should be passing in an array initializer with the CVar value tied to Skip Jiggy Dance, since this shouldn't be necessary without it. Of course, these statements apply to any other place you added ShipInit and listeners.

src/port/patches/FramePacingPatches.cpp
```diff
@@ -0,0 +1,136 @@
...
+} // extern "C"
+
+void RegisterFramePacingPatches_Init() {
+    REGISTER_LISTENER(OnMapLoad, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        (void)event;
+        resetCutsceneTimings();
+    });
+}
+
+static RegisterShipInitFunc initFunc(RegisterFramePacingPatches_Init);
```
Malkierian, line 144: The entirety of this file's functionality, wherever it's inserted, should be gated behind a Fix enhancement that we can probably default to on, as well as a corresponding VanillaBehavior.

src/port/patches/Patches.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Input
+
+float port_getRumbleScale(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
Malkierian, line 56: To be honest, I think the vast majority of this file should be possible to eliminate if events are properly implemented, but they don't all have to be handled in this PR if they're not added in this PR.

src/port/patches/SavePatches.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+void RegisterSavePatches_Init() {
+    REGISTER_LISTENER(OnFileLoaded, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        OnFileLoaded* ev = (OnFileLoaded*)event;
+        onFileLoaded(ev->gamenum, ev->filenum);
+    });
+}
+
+static RegisterShipInitFunc initFunc(RegisterSavePatches_Init);
```
Malkierian: I get the feeling this whole file (and the correlating places that were pulled out of in decomp code) will become redundant, if not a source of conflicts, with Cal's upcoming save/load flow rework, but I'm not sure if that should keep this out.

src/port/patches/SpritePatches.cpp
```diff
@@ -0,0 +1,54 @@
...
+        sSpriteDisplayCache[sSpriteDisplayCacheCount].displayData = dd;
+        sSpriteDisplayCacheCount++;
+        return dd;
+    }
+    func_803382E4(-1);
+    func_80338308(sprite_getUnk8(sprite), sprite_getUnkA(sprite));
+    return func_80344A1C(sprite);
+}
+
+} // extern "C"
```
Malkierian, line 54: I want to say this could be hooked, but I'm not sure if it should be.

Malkierian (changes requested):
A good start, just some lack of understanding about the way events and ShipInit interact, as well as our stance on "always-on" changes vs "default-on" enhancements.

src/core1/io/pfs_manager.c
JeodC: This is the solution to attract demo desync in stop n swop and bottles bonus demos. It comes from recomp and when testing the sns pictures ending for all 100 jiggies at the normal viCount of 2, the demos ended before the dialog did, cutting off the dialog. Setting viCount to 3 for these specific instances allows the cutscene and dialog to synchronize properly. By not setting it, we're actively avoiding VB.

As for why it's in pfs_manager.c, it's because demo input logic is performed here.

src/core2/anim/anim_bonetransform.c
JeodC: Yeah I am okay with doing that.

src/port/patches/CameraPatches.cpp
JeodC: Need to be selective with this file I suppose, I personally feel that supporting widescreen this way isn't something to be gated behind an enhancement or a fix because when not enabled it looks like negligence or apathy on our part. Default enable, sure, but then who would ever disable it?

src/port/patches/FramePacingPatches.cpp
JeodC: See above re: defining VB. I had been defining VB as something to always keep, and fix enhancements to be an adjustment to VB bugs such as 1.0 -> 1.1 fixes and other bugs mentioned on various websites.

src/port/patches/SavePatches.cpp
JeodC: It may, but I can only work with what's currently available without being dependent on a future pull request.

src/port/patches/SpritePatches.cpp
JeodC: This replaces broken rom cache behavior that doesn't work on PC — without it sprites render incorrectly. It's structural, not behavioral.

src/port/patches/Patches.h
JeodC: Future work, this file still exists for patches that don't cleanly translate to event hooks. The ones that *did* translate cleanly are decision points: "a track is being set", "a map is loading", "a static camera activated", etc. The functions in this file don't fit into the event hooks model well.

These functions return values that the decomp code needs immediately in an expression:

- `port_consumeReadbackRequest()` — returns 0/1, used in an if condition in the display list builder
- port_getPauseFramebufferId() — returns a framebuffer id used directly in a gfx command, and is a structural port function, not behavioral
- `port_getAuxGpuFbId()`, `port_getTransitionGpuFbId()` — same, returns GPU FB ids, also structural
- `port_shouldCaptureTransition()` — returns bool, used in if guard
- `port_getRumbleScale()` — returns a float used directly in a multiplication — it reads the controller config, not a behavioral choice
- `port_getCutsceneExtraVis()` — returns int added to a VI count

These are called per-frame from rendering or audio paths (has overhead):

- `port_readAuxFbToCpu()`, `port_readTransitionFbToCpu()` — emit gfx commands into the display list, structural addition needed for bottles bonus and falling jiggy transition framebuffers
- `port_patchPictureModel()`, `port_patchTransitionModel()` — walk and rewrite model display lists, structural addition needed to prevent visible seams in these models from the individual jigsaw pieces
- `port_spriteDisplayCache_clear()` — called during model cache flush, also structural because without it the game breaks rendering sprites

No behavioral choice to override:

- `port_setViBlack()`, `port_freezeReadback()`, `port_requestReadback()` — set/clear a flag
- `port_setDemoViCount()` — set a value
- `port_camera_applyWsYawFix()` — per-frame rotation adjustment, runs per frame as long as the camera is active

src/port/patches/CameraPatches.cpp
JeodC: COND_HOOK doesn't exist in Lighthouse yet. Happy to swap to it when it's added. I added COND_VB_SHOULD but I don't think it's correct usage here. These listeners track camera state unconditionally — there's no CVar to gate on. COND_VB_SHOULD with `true` adds de-registration overhead for no benefit.

Malkierian: Leave a TODO here, then, and on any other non-VB ShipInit hook, as I'll try adding that in this weekend and will swap them myself.

JeodC: Done.

src/core1/io/pfs_manager.c
Malkierian: No, I just mean the function instead of direct assignment. The function literally just assigns it the same way anyway, so it's redundant.

src/port/patches/FramePacingPatches.cpp
Malkierian: VB and Fixes are neither inherently the same nor mutually exclusive. Many Fixes in SoH use VBs, because VBs are things we want to branch only if the user wants (or we want it default on), but should use vanilla code when we're not branching.

src/port/patches/CameraPatches.cpp
Malkierian: Yeah, if there are widescreen changes and things for jiggy dance mixed in here, they should be separated into different hook processes. Also, there are plenty of people who want the original aspect ratio experience for these ports, and unless you can absolutely guarantee that the way the widescreen adaptation works replicates vanilla when in original aspect, even if it should be auto-on (which I agree with), a way to turn it off should be created.

src/port/patches/Patches.h
Malkierian: Cool, I just wanted to make the point that eventually this should be minimized as much as possible. I realize that probably not all of them could be.

src/core1/io/pfs_manager.c
JeodC: The setter is needed because the variable moved out of pfs_manager.c into FramePacingPatches.cpp — direct assignment is no longer possible. The display count wrapper keeps the map-specific overrides in port code. We could inline it again, but that means adding map-specific VI overrides to the decomp file.

src/port/patches/CameraPatches.cpp
JeodC: The jiggy dance skip isn't in CameraPatches. This file only contains the cutscene aspect lock and the widescreen yaw fix.

PR #41, opened 2026-04-02
Reimplement Lair Continuity using the event system

JeodC:
- Replace direct extern calls from `audio_instruments.c` with `VB_MUSIC_*` events (`SET_TRACK`, `TICK`, `PRE_STOP`, `TRACK_START`)
- LairContinuity.cpp registers as an event listener instead of being called directly from decomp
- Seek functions (`lairAudio_consumePendingSeek`, `lairAudio_hasPendingSeek`) remain as direct calls via Patches.h since they return data from `n_csplayer.c` and `audio_musicplayer.c`
- Add `AudioAccessors.c` for thin C accessors to decomp audio structs, avoiding header conflicts with C++
Decomp file (audio_instruments.c) no longer has any lair continuity externs or inline logic

Depends on #40

### DIFF
--- a/src/core1/audio_instruments.c
+++ b/src/core1/audio_instruments.c
@@ -32,39 +32,8 @@ void func_8024FD28(u8, s16);
 int func_80250074(u8);
 u8 func_8025F4A0(ALCSPlayer *, u8);
 
-// [port] Lair audio continuity: tracks within the same group share enough
-// sequence data that seeking to the same tick position sounds seamless.
-// Group 0 = not a lair track, groups 1-3 = adjacent lair floors.
-static int lairTrackGroup(s32 trackIndex) {
-    switch (trackIndex) {
-        case COMUSIC_1E_GL_MM_VERSION:
-        case COMUSIC_50_GL_TTC_VERSION:
-        case COMUSIC_51_GL_CCW_VERSION:
-            return 1;
-        case COMUSIC_52_GL_BGS_RBB_VERSION:
-        case COMUSIC_53_GL_FP_VERSION_A:
-            return 2;
-        case COMUSIC_54_GL_GV_VERSION:
-        case COMUSIC_59_GL_FP_VERSION_B:
-        case COMUSIC_5D_GL_MMM_VERSION:
-        case COMUSIC_5E_GL_MMM_RBB_VERSION:
-        case COMUSIC_63_GL_FF_VERSION:
-            return 3;
-        default:
-            return 0;
-    }
-}
-
-// Per-group saved tick positions, plus a combined position for cross-group seeks.
-static s32 sLairSavedTicks[4] = { 0 }; // one per group (index 1-3)
-static s32 sLairLastSavedTicks = 0;     // last saved from any group
-static s32 sLairPendingSeek = 0;        // ticks to seek to on next track start
-static u8  sLairPendingSeekSlot = 0xFF; // which slot the seek is for
-
-// [port] Defined after D_80281720 declaration
-s32 lairAudio_consumePendingSeek(ALCSPlayer *player);
-// [port] Returns true if a lair continuity seek is pending (for skipping fade-in)
-int lairAudio_hasPendingSeek(void) { return sLairPendingSeek > 0; }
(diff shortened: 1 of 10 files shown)


src/port/enhancements/LairContinuity.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return ticks;
+}
+
+// Called directly from audio_musicplayer.c via Patches.h
+extern "C" int lairAudio_hasPendingSeek(void) {
+    return sPendingSeek > 0;
+}
+
+void RegisterLairContinuity_Init() {
+    REGISTER_LISTENER(VanillaBehavior, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
```
Malkierian: Change to COND_VB_SHOULD.

src/core1/audio_instruments.c
```diff
@@ -404,15 +362,11 @@ void func_8024FA98(u8 arg0, s32 arg1){
...
-            int newGroup = lairTrackGroup(arg1);
-            if (newGroup > 0 && sLairLastSavedTicks > 0) {
-                sLairPendingSeek = sLairSavedTicks[newGroup] ? sLairSavedTicks[newGroup] : sLairLastSavedTicks;
-                sLairPendingSeekSlot = arg0;
-            }
+        // [port] Lair continuity: set up pending seek for lair-to-lair transitions
+        if (sp2C == -1 && arg1 >= 0) {
+            bool should = true;
+            s32 trackArgs[2] = { arg0, arg1 };
+            CALL_EVENT(VanillaBehavior, VB_MUSIC_TRACK_START, &should, trackArgs);
```
Malkierian: I think this should be a cancellable event, and it looks like `should` is unused and should be removed.

src/core1/audio_instruments.c
```diff
@@ -450,43 +404,13 @@ void func_8024FB8C(void){
...
-            D_80281720[arg0].unk3 = 0;
-            D_80281720[arg0].unk0 = 0;
-        }
-    }
+    // [port] Lair continuity hook — may defer the stop or set up a seek
+    bool should = true;
+    s32 trackArgs[2] = { arg0, arg1 };
+    CALL_EVENT(VanillaBehavior, VB_MUSIC_SET_TRACK, &should, trackArgs);
+    if (!should)
+        return;
```
Malkierian, line 411: Should also make this into a cancellable event. For this one, you'd just wrap the entirety of the rest of the function in curly brackets, then set cancelled in the listener, rather than this early return. You can use the OnActorSpawn call in `actor_cubebounds.c` as reference.

src/core1/audio_instruments.c
```diff
@@ -560,31 +484,17 @@ s32 func_8024FEEC(u8 arg0)
 
 void func_8024FF34(void){
     s32 i;
+    bool should = true;
```
Malkierian: Also should be removed.

Malkierian (changes requested):

src/core1/audio_instruments.c
JeodC: So `OnActorSpawn` is a custom event with its own `IEvent` header, so cancelling via `event->cancelled = true` works there. But `VanillaBehavior` carries a `bool* should` field that existing `COND_VB_SHOULD` listeners (e.g. `CameraPatches.cpp`) use via `*ev->should = false`.

If I switch the call site to `CALL_CANCELLABLE_EVENT` and keep `COND_VB_SHOULD` on the listener side, the two mechanisms don't line up. So what should actually be done here?

Malkierian: Ah, if VB's are cancellable via setting the internal should to false, that should be fine. At some point we should make a macro or wrapping function to make that easier. In the end, though, VB should only be used if there's vanilla code we want to maintain as a possible execution path. As I was evaluating this, I looked into your previous uses of VB in staticCamera, and those don't really qualify as VB.

JeodC:
I found and fixed an issue where `COND_VB_SHOULD`'s first unregister call would destroy all other listeners.

PR #43, opened 2026-04-02
Hook up quit/reset buttons, improve game stats, fix UI strings

JeodC:
- Add reset and quit console commands so the imgui menu buttons work
- Introduce `gPortResetPending` flag to break audio spin-waits that deadlock when the audio callback isn't running during reset
- Improve window title stats: show "FILE SELECT" label, hide stats during cutscenes and file select, use "--" placeholders
- Fix "Quit SoH" to "Quit Lighthouse" in quit confirmation dialog
- Move map change requirement into dialog language widget label and expand tooltip
- Clean up ShipUtils: consolidate extern "C" block, remove `port_getViewportWidth`, tidy comments


JeodC:
@Malkierian Rebased and ready for review.

PR #54, opened 2026-04-06
Enhancement: Faster swimming + enhancement cleanup

### LINKED ISSUE #48: [ENHANCEMENT] Faster swimming
In Banjo-Tooie the player can save Roystein the goldfish in Spiral Mountain to unlock faster swimming where holding A and B at the same time causes both Banjo and Kazooie to work together to swim faster. This could be backported as an enhancement, if the following conditions can be met:

- We can play two animations (Banjo's slower swim and Kazooie's faster swim) at the same time
- We can adjust the swim speed conditionally to add both velocity values together or trigger both at the same time
- The game can read and accept two inputs at the same time (it should)

JeodC:
Backport from Banjo-Tooie and use a trick to have two animations playing at once, thanks Bottles!

Will resolve #48 

Also cleans up my prior enhancements using pre-hook system.


src/core2/bs/player_spawn.c
```diff
@@ -920,12 +920,17 @@ void func_8029CBC4(void){
...
         if( jiggyscore_total() == 100 && fileProgressFlag_get(FILEPROG_FC_DEFEAT_GRUNTY)){
             timedFunc_set_3(4.1f, (GenFunction_3)transitionToMap, MAP_95_CS_END_ALL_100, 0, 1);
         }//L8029CC58
-
-        timedFunc_set_0(4.0f, func_8029CBC4);
+        if (shouldDance) {
+            timedFunc_set_0(4.0f, func_8029CBC4);
+        } else {
+            func_8029CBC4();
+        }
```
Malkierian, line 931: I think you can move the should here directly instead of storing it in a bool at the start of the function.

src/core2/gamestate.c
```diff
@@ -76,7 +76,7 @@ s32 item_adjustByDiff(enum item_e item, s32 diff, s32 no_hud){
             || (item == ITEM_F_RED_FEATHER && volatileFlag_get(VOLATILE_FLAG_75_SANDCASTLE_INFINITE_RED_FEATHERS))
             || (item == ITEM_10_GOLD_FEATHER && volatileFlag_get(VOLATILE_FLAG_76_SANDCASTLE_INFINITE_GOLD_FEATHERS))
             || (item == ITEM_17_AIR && volatileFlag_get(VOLATILE_FLAG_96_SANDCASTLE_INFINITE_AIR))
-            || (item == ITEM_14_HEALTH && CVarGetInteger(CVAR_DEVELOPER_TOOLS("InfiniteHealth"), 0))
+            || (item == ITEM_14_HEALTH && port_isInfiniteHealth())
```
Malkierian: This could be rolled into an OnItemPickup or OnPickupProcess event, but doesn't have to be now.

src/FP/ch/boggy2.c
```diff
@@ -340,9 +340,8 @@ void func_803893E4(Actor *this, f32 arg1, u8 arg2){
         sp2C = 1100.0f;
```
Malkierian, line 340: Is this function basically `boggyRace_init` like it looks like? If so, it might be worth making an overall hook for the end of the function that then modifies the speed. Could send the actor pointer then.

src/port/enhancements/cutscenes/CutsceneSkips.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+void RegisterSkipIntroCutscene_Init() {
+    COND_HOOK(OnIntroCutsceneCheck, EVENT_PRIORITY_NORMAL, CVarGetInteger(CVAR_SKIP_INTRO, 0), [](IEvent* event) {
+        auto* ev = reinterpret_cast<OnIntroCutsceneCheck*>(event);
+        *ev->skipIntro = true;
+    });
+}
+
+void RegisterSkipJiggyDance_Init() {
+    COND_VB_SHOULD(VB_SKIP_JIGGY_DANCE, EVENT_PRIORITY_NORMAL, CVarGetInteger(CVAR_SKIP_JIGGY_DANCE, 0),
```
Malkierian: Considering the nature of the `should` here being true by default, perhaps the VB should be named `VB_PLAY_JIGGY_DANCE`, as false then reflects the true nature of the interaction at the other end of the event.

src/FP/ch/boggy2.c
JeodC: This is a per-frame speed update, not init.

src/core2/gamestate.c
JeodC: I actually think this is cleaner if driven by the TTC cheat.

Malkierian (approved):

PR #55, opened 2026-04-07
Fix loading vtx section before gfx in model import and add sentinel ENDDL byte

JeodC:
This fixes a crash in Nostalgia64 and is correct for N64 behavior, where GFX comes before VTX. We add a sentinel byte because some models exported by Torch don't have an ENDDL byte, so the importer reads past the DL into the next chunk. On N64, this would be the VTX section which is harmless. If GFX is the final chunk, though, we read into heap and can crash. But we also add a sentinel byte so we don't shoot past the end of the DL in the first place.

Also fixes a bug in importing move flags from romhacks.

Also fixes save isolation which was lost in the transition to the new save system.

### DIFF
--- a/src/port/save/SaveManager.cpp
+++ b/src/port/save/SaveManager.cpp
@@ -3,6 +3,7 @@
 #include <libultraship/bridge/consolevariablebridge.h>
 #include "port/enhancements/events/hooks/Events.h"
 #include "port/ShipUtils.h"
+#include "port/GameConfig.h"
 #include <fstream>
 #include <filesystem>
 #include <regex>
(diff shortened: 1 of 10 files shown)


src/port/save/SaveManager.cpp
```diff
@@ -24,6 +25,16 @@ using nlohmann::ordered_json;
 namespace fs = std::filesystem;
 static bool mLoaded = false;
 
+std::string SaveManager_GetSavePath(const std::string& filename) {
+    const char* romName = port_getRomhackName();
+    if (romName && romName[0] != '\0') {
```
Malkierian: ShipUtils has a CStringEmpty function you can use here instead, I think.

Malkierian (approved):

PR #56, opened 2026-04-07
Build/warning fixes

JeodC:
This fixes ~1,100 warnings across port cpp files and header files. Decomp c files are for the most part untouched.


src/port/ui/Notification.cpp
```diff
@@ -41,9 +41,9 @@ void Window::Draw() {
     ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, ImVec2(8.0f * CVarGetFloat("gNotifications.Size", 1.8f), 6.0f));
     ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(8.0f * CVarGetFloat("gNotifications.Size", 1.8f), 8.0f));
 
-    for (int index = 0; index < notifications.size(); ++index) {
+    for (int index = 0; index < (int)notifications.size(); ++index) {
```
Malkierian: Since it's iterating on a size_t, might as well make index size_t as well, and handle conversions after that. Should still resolve the warning.

JeodC: Done.

Malkierian (approved):

PR #87, opened 2026-04-15
Sync with decomp upstream: header extraction, file reorg, boot stubs

JeodC:
Decomp-pass-3: a three-round audit of Lighthouse's sync with upstream `banjo-decomp` (HEAD `d25b223`) from the baseline commit `ab1913ed`, followed by the layout cleanup upstream done alongside its recent header-extraction work.

Four independent audit passes found **55+ missed renames and mismatches** across the eleven commits of pass 2, all applied here. On top of that, the header/file layout is brought in line with upstream's reorganization: new `core2/*` headers, audio/PR
headers moved to `lib/libultra/`, libaudio's `n_*` variant split out to `n_audio/`, and boot files relocated to `src/boot/`.


src/boot/rarezip.c
```diff
@@ -52,9 +52,9 @@ s32 D_80275768 = 6; //dbits
 
 /* .data */
 struct huft D_803FBE00;
-struct huft *D_8027BF00;
+struct huft *inflate_huft;
 u8 pad_8027BF08[0x8];
-u8 *inflate_huft; //inbuf
+u8 *inflate_inbuf; //inbuf
 u8 *D_8027BF14; //slide
 u32 inflate_inptr; //inptr
```
Malkierian: The comments for the now-named variables can probably be removed? Are they still in decomp?

src/core1/audio_manager.c
```diff
@@ -4,9 +4,9 @@
 #include "functions.h"
 #include "variables.h"
 #include "version.h"
-//#include "2.0L/PR/sched.h"
-#include "2.0L/PR/n_libaudio.h"
-//#include "2.0L/PR/os_system.h"
+//#include <PR/sched.h>
+#include <n_audio/PR/n_libaudio.h>
```
Malkierian: <> should only be used for system and submodule includes. Since libaudio is part of the port code, it should still be in "".

src/boot/rarezip.c
JeodC: Needs to be amended in pass 4, rarezip.c no longer exists upstream and these are now in rarezip.h. Ok to keep for this round?

src/core1/audio_manager.c
JeodC: Will resolve when I find some time this weekend

PR #89, opened 2026-04-18
[ADD] - Lots of cheat

PurpleHato:
Lots of new cheats + proper separator if we continue to follow that path of having one file containing everything

### DIFF
--- a/src/port/enhancements/cheats/Cheats.cpp
+++ b/src/port/enhancements/cheats/Cheats.cpp
@@ -9,10 +9,58 @@
 
 extern "C" {
 #include "enums.h"
+#include "core2/statetimer.h"
+#include "bs_funcs.h"
 void volatileFlag_set(enum volatile_flags_e index, s32 set);
+void item_setMaxCount(s32 item);
+void item_set(s32 item, s32 val);
+s32 item_getCount(enum item_e item);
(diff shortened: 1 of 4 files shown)


JeodC (approved):
LGTM

src/port/enhancements/cheats/Cheats.cpp
```diff
@@ -9,10 +9,59 @@
...
+void batimer_set(s32 timer_id, f32 duration);
+f32 batimer_get(s32 timer_id);
+s32 batimer_decrement(s32 timer_id);
+void batimer_incrementBy(s32 id, f32 inc_value_sec);
+s32 baflag_isTrue(s32 flag);
+f32 time_getDelta(void);
+s32 getGameMode(void);
+void baphysics_set_gravity(f32 gravity);
+void baphysics_reset_gravity(void);
+enum map_e gsworld_getMap(void);
```
Malkierian: It would probably be better to add these to `functions.h` if they're not in there already, since I'm sure a good number of them will be used by other enhancements in the future. Just follow the current format for listing what source file they're coming from.

PurpleHato: Done, there was indeed many that I could have cleaned up and added the one you asked o7

PR #108, opened 2026-04-29
Add proper mod menu and bump Torch so romhacks are generated as mod o2r files

JeodC:
- Ship with new `hashes.yml` file
- Torch uses this file as a diff against romhacks to only extract what's unique
- `aGameConfig` is still generated and logic is kept intact
- Add mod menu improvements from Shipwright and add Disabled Mods menu
- Add boot check to disable all romhacks if multiple are enabled and show a warning popup while booting vanilla `bk.o2r` and non-romhack mods

Torch detects BB romhacks properly by:
- Quick routing check: ROM size > 16MB (vanilla v1.0/v1.1/PAL are all 16MB; BB-extended ROMs are 64MB)
- Fingerprint validation: `FindPatchedOverlay` walks 9 candidate EOR bootstrap pointer pairs, derives an address from each, and looks for a BKZIP magic (0x11 0x72) followed by a sane decompressed length (800KB–1MB). If `bk_unzip` produces a buffer matching the declared size, the ROM is confirmed BB.
- An extended ROM that passes the size check but fails the overlay scan is rejected — Torch refuses to synthesize a BK64 config rather than attempting a doomed extraction on a non-BK ROM.

Additional:
- Torch inserts SHA1 hashes into `aGameConfig` for both the rom itself and for any custom code blobs it finds
- Lighthouse now has a system to identify romhacks by sha1 table and can define special behaviors for each
- Proof of concept in Banjo Dreamie


src/core2/sprite/rendertex.c
```diff
@@ -268,7 +268,7 @@ void func_80348044(Gfx **gfx, BKSprite* sprite, s32 frame, s32 tmem, s32 rtile,
             D_80386098 = (s32) D_80386098 >> 1;
         }
     } else if (sprite->type & SPRITE_TYPE_CI4) {
-        rare_gDPLoadMultiBlock_4b((*gfx)++, timg, tmem, rtile, G_IM_FMT_RGBA, *width, *height, uls, ult, 0, cms, cmt, masks, maskt, G_TX_NOLOD, G_TX_NOLOD);
+        rare_gDPLoadMultiBlock_4b((*gfx)++, timg, tmem, rtile, G_IM_FMT_CI, *width, *height, uls, ult, 0, cms, cmt, masks, maskt, G_TX_NOLOD, G_TX_NOLOD);
```
JeodC, line 271: Note: this is a decomp error which caused hud sprites to be rendered incorrectly, which lus at one point fixed but ended up masking the cause

Malkierian:
Would like to see, if possible, that we follow the Starship flow of extracting voiceover sets from other ROMs applied to hack o2r creation. This is done via menu buttons rather than swapping ROMs at boot.

JeodC:
> Would like to see, if possible, that we follow the Starship flow of extracting voiceover sets from other ROMs applied to hack o2r creation. This is done via menu buttons rather than swapping ROMs at boot.

This is integrated.

Issue #107, opened 2026-04-29
[ENHANCEMENT] Honeyback with all empty honeycombs


JeodC:
There are 24 empty honeycombs in Banjo-Kazooie, but only 18 award a new health slot. This is because a 9th health slot will overlap other HUD elements unless widescreen is used.

Instead, backport the Honeyback health regen over time feature from Banjo-Tooie as an optional enhancement and grant it when all 24 empty honeycombs are collected.

Issue #78, opened 2026-04-12
[ENHANCEMENT] Disable all LOD


JeodC:
The `AlwaysHighPolyBanjo` enhancement should be extended to disable all LOD. LOD in models is more noticeable in ultrawide, specifically in TTC where the palm trees, flight pads, and shock spring pads all gain higher LOD as they are approached.

Issue #74, opened 2026-04-08
[FIX] Grunty Fight Jinjo Sound Bug


JeodC:
From TCRF:

> In the Grunty Fight, there are Jinjos the player must unlock from statues that hit Grunty to progress the fight. On US v1.0, if Banjo hits Grunty with a Jinjo before they finish their charge up animation (can be done by guiding the Jinjo into her because it spirals above Banjo), the charge up sound effect will annoyingly play for the rest of the fight. This was patched on European, US v1.1, Japanese, and XBLA as now the sound effect concludes as soon as the Jinjo hits Grunty from any point in their animation.

Marked in code [here](https://github.com/HarbourMasters/Lighthouse/blob/0da36486a0b4dd9a9378c05e6eb5867dd3386653/src/FINALE/ch/jinjo.c#L283) and just needs conversion to event hook and test.

Issue #73, opened 2026-04-08
[FIX] Boggy Race Game Over


JeodC:
From TCRF:

> In Freezeezy Peak, the player can race Boggy the Polar Bear as the Walrus transformation and as Banjo & Kazooie with the speed shoes for a second time. On US v1.0 and European, if the player were to lose the 1st or 2nd race with 0 lives, the game would initiate a game over. On US v1.1, Japanese, and XBLA, losing the Boggy Race with 0 lives no longer gives the player a game over, as it teleports the player back to the start of the race. The other minigames such as TTC Sandcastle, GV Matching Pyramid, and MMM Tumblar still initiate a game over at 0 lives on all versions.

Marked in code [here](https://github.com/HarbourMasters/Lighthouse/blob/0da36486a0b4dd9a9378c05e6eb5867dd3386653/src/core2/bs/player_spawn.c#L279) and just needs to be converted to event hook and then tested.

PR #135, opened 2026-06-03
Bring in more fixes and enhancements

JeodC:
Addresses the following issues:

#65, #66, #67, #68, #70, #72, #73, #74, #78, and #107

The remaining are deferred. Model collision is asset-side and needs a team plan of approach if we want to have code-side fixes. Pushing and bouncing Grunty in the final fight had a marker that on further analysis was a no-op, so it was removed and I have not been able to trace the proper collision path for Grunty's phase 4.

### DIFF
--- a/src/CCW/flower.c
+++ b/src/CCW/flower.c
@@ -98,12 +98,10 @@ void CCW_func_80388278(ActorMarker *marker, ActorMarker *other_marker) {
 void func_803882A4(ActorMarker* marker, ActorMarker *other_marker) {
     Actor* actor = marker_getActor(marker);
 
-#ifdef PORT_FIX
     // [port] v1.1 fix: don't allow re-planting flower in Spring after it's already been planted
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 && !fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING)) {
-#else
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1) {
-#endif
+    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 &&
+        (!fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING) ||
+         EventSystem_Should(VB_CCW_FLOWER_REPLANT, true))) {
         func_80387F64(actor, 2);
     }
 }
--- a/src/CCW/flower.c
+++ b/src/CCW/flower.c
@@ -99,9 +99,9 @@ void func_803882A4(ActorMarker* marker, ActorMarker *other_marker) {
     Actor* actor = marker_getActor(marker);
 
     // [port] v1.1 fix: don't allow re-planting flower in Spring after it's already been planted
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 &&
-        (!fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING) ||
-         EventSystem_Should(VB_CCW_FLOWER_REPLANT, true))) {
+    if (EventSystem_Should(VB_CCW_FLOWER_REPLANT,
+                           gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 &&
+                               !fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING))) {
         func_80387F64(actor, 2);
     }
 }
--- a/src/port/patches/GraphicsPatches.cpp
+++ b/src/port/patches/GraphicsPatches.cpp
@@ -14,7 +14,7 @@ int port_getDrawDistanceLevel(void) {
     return level;
 }
 
-int port_shouldForceHighPolyBanjo(void) {
-    return CVarGetInteger(CVAR_ENHANCEMENT("Graphics.AlwaysHighPolyBanjo"), 0);
+int port_shouldDisableLOD(void) {
+    return CVarGetInteger(CVAR_ENHANCEMENT("Graphics.DisableLOD"), 0);
 }
 }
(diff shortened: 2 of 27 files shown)


src/CCW/flower.c
```diff
@@ -98,12 +98,10 @@ void CCW_func_80388278(ActorMarker *marker, ActorMarker *other_marker) {
...
     Actor* actor = marker_getActor(marker);
 
-#ifdef PORT_FIX
     // [port] v1.1 fix: don't allow re-planting flower in Spring after it's already been planted
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 && !fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING)) {
-#else
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1) {
-#endif
+    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 &&
+        (!fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING) ||
```
Malkierian: Use the `fileProgressFlag_get` call here inside the _Should, instead of the true. Same goes for all other calls to `EventSystem_Should()`.

src/port/patches/GraphicsPatches.cpp
```diff
@@ -14,7 +14,7 @@ int port_getDrawDistanceLevel(void) {
     return level;
 }
 
-int port_shouldForceHighPolyBanjo(void) {
-    return CVarGetInteger(CVAR_ENHANCEMENT("Graphics.AlwaysHighPolyBanjo"), 0);
+int port_shouldDisableLOD(void) {
```
Malkierian, line 17: All of these port_ functions should likely eventually be changed to EventSystem_Should calls.

src/port/ui/LighthouseMenuEnhancements.cpp
```diff
@@ -132,6 +168,12 @@ void LighthouseMenu::AddMenuEnhancements() {
         .Options(CheckboxOptions().Tooltip("Adjusts static camera angles in widescreen to prevent skybox "
                                            "exposure at the edges of the screen."));
 
+    AddWidget(path, "Center Enemy SFX", WIDGET_CVAR_CHECKBOX)
```
Malkierian, line 173: Center Enemy SFX, Boggy Race Game Over, and CCW Flower Softlock should probably all have defaults of on. These defaults will need to be handled in the VB handlers as well.

src/CCW/flower.c
JeodC: That makes the result encode the fixed behavior by default, so the VB handler would flip to "restore vanilla when the CVar is off" (`*should = true`) rather than the usual "apply fix when on" (`*should = false`) that every other fix uses. Is that the intent, or did you want a different handler shape? And for "same goes for all other calls" — do you mean only the ones with a real value to fold in (this is the only such case; the rest pass a plain true/false), or something broader?

src/port/patches/GraphicsPatches.cpp
JeodC: Do you want these converted in this PR, or tracked for a follow-up? Two specifics if now:

- `port_getDrawDistanceLevel` returns an int level (0–4) plus an attract-demo override, so it doesn't fit the bool `EventSystem_Should` gate. Would that need a different (int-returning) event, or stay `port_`?
- `port_shouldDisableLOD` is bool and converts cleanly, but it's called per-geo-command in `modelRender_geoCmd_LOD`, so a raw `EventSystem_Should` there means an event dispatch in that hot loop every frame. In short, it's designed this way to mitigate costs, but I can try caching it. The _Should would still fire every frame, but the code would use the cached result.

src/port/ui/LighthouseMenuEnhancements.cpp
JeodC: The CCW Flower one depends on how the handler shape lands in the `flower.c` thread above, rest are easily flipped.

src/CCW/flower.c
Malkierian: Oh, I'm not sure how I missed it, but I meant to do the whole condition for the Should condition. The idea is that VBs can go either way depending on how you word it. The point is that the condition for Shoulds should only be true if the original condition is true, which causes the vanilla code to execute. Shoulds should return false for not wanting to execute vanilla code.

As for the "all other calls", I did mean for when there's a real condition to fold into it.

src/port/patches/GraphicsPatches.cpp
Malkierian: It can be done in a separate PR, I know there are a lot of them, and they're perfectly functional as is. Just wanted to maintain complete consistency across fixes/enhancements. You can pass pointers into VBs that the handlers can then pull out of va_args and then modify by dreferencing.

As for performance concerns, the event system's overhead should be negligible. You have to have hundreds of thousands of calls to resizable container reading/modifying (like the alt assets CVar before I swapped it out for a variable) before they have a noticeable effect on performance, and even then, it will generally only happen in debug builds or on very low-end hardware.

src/port/ui/LighthouseMenuEnhancements.cpp
Malkierian: They shouldn't need any inversions, as long as the handling reflects the implication of the naming. The off states for fix and enhancement checkboxes should always be vanilla behavior. if there's ever an instance where we would want to be able to explicitly apply specific states, then we'd have a combobox that would say "default (unmodified", "always v1", and "always v2". I think there's an example of this in SoH, but I can't remember exactly what it was off the top of my head.

src/CCW/flower.c
```diff
@@ -98,12 +98,9 @@ void CCW_func_80388278(ActorMarker *marker, ActorMarker *other_marker) {
 void func_803882A4(ActorMarker* marker, ActorMarker *other_marker) {
     Actor* actor = marker_getActor(marker);
 
-#ifdef PORT_FIX
     // [port] v1.1 fix: don't allow re-planting flower in Spring after it's already been planted
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 && !fileProgressFlag_get(FILEPROG_E3_CCW_FLOWER_SPRING)) {
-#else
-    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1) {
-#endif
+    if (gsworld_getMap() == MAP_43_CCW_SPRING && actor->state == 1 &&
```
Malkierian: This should also be included in Should condition

PR #138, opened 2026-06-06
JP: Fix a compile issue for strtok, add JP features

JeodC:
- Fix strtok conflict in building
- Fix JP dialog rendering crash
- Move hardcoded strings to a new LocalizedText.cpp file (decomp based on US v1.0)
- Integrate special rendering for pause menu banners in JP
- Integrate subtitles for character parade when using JP
- Bump LUS for a crash fix when rendering pause menu banners
- Bump Torch for extracting the Kana sprites as sprites instead of blobs


src/port/ui/Menu.cpp
```diff
@@ -98,16 +98,17 @@ void Menu::RemoveSidebarSearch() {
 }
 
 void Menu::UpdateWindowBackendObjects() {
-    Ship::WindowBackend runningWindowBackend = Ship::Context::GetInstance()->GetWindow()->GetWindowBackend();
+    int32_t runningWindowBackendId = Ship::Context::GetInstance()->GetWindow()->GetWindowBackend();
```
Malkierian, line 101: What was the benefit of this here, instead of keeping it as WindowBackend?

JeodC: This is from the LUS bump. `Ship::WindowBackend` no longer exists. `Window::GetWindowBackend()` now returns `int32_t` and `GetAvailableWindowBackends()` now returns `std::shared_ptr<std::vector<int32_t>>`.

Malkierian: I knew Ship::WindowBackend was gone, but Fast::WindowBackend replaced it. weird that they would change the return type to remove the enum like that, though.

PR #147, opened 2026-06-14
Alt assets

JeodC:
Depends:
#140 
https://github.com/Kenix3/libultraship/pull/1118 (currently merged to `port-maintenance`)

Some notes:
- Torch bump was necessary because BK models didn't carry a sentinel to signal alt texture paths; all o2r files will need regeneration.
- Localized alt asset differences have a special path following lang-picker. When using a single `bk-reloaded.o2r`, for example, it will load `alt\assets\lang\jp\model\` when Japanese is chosen, instead of `alt\assets\model\`. In this case, the title sign is different in JP.
- HUD Jinjo heads are a single asset (7DF) in decomp and are colored at runtime. Alt assets may ship pre-colored heads, so https://github.com/HarbourMasters/Lighthouse/pull/147/commits/cfe8642a9213ad40f9d55807e8105dd967649a1b accounts for this. Alt assets can still ship a single 7DF if they want the runtime coloring.
- The bold font used in HUD is a product of `mask * sphere`, calculated at runtime. Similar to Jinjo heads, some alt asset packs may ship the final product, which normally isn't a referenced static asset. https://github.com/HarbourMasters/Lighthouse/pull/147/commits/1e3646a1b88714960401ef1ccc4855db0be99597 enables the option to either use the final output in alt assets or use the two components and allow the port to calculate the end result.

As of https://github.com/HarbourMasters/Lighthouse/pull/147/commits/a9f968a309501b2e18843d64324c8634c5029c71 romhacks are given their own menu, and Mod Menu is contextualized to only display mods compatible with the selected romhack. Compatibility is driven by file location, not by o2r content. User responsibility.

image

image


Malkierian:
I assume this has been updated to latest commit on port-maintenance?

Malkierian:
Also, what's this sentinel you speak of?

JeodC:
> Also, what's this sentinel you speak of?

Torch stamps `0xFF` in the high byte of each static `G_SETTIMG`'s w1 word and packs the texture index into the low 24 bits; Lighthouse treats that `0xFF` as a "this is a replaceable texture #N" sentinel and routes it through the alt-asset path instead of the raw embedded image.

This isn't how sister ports do it. In `ModelFactory.cpp#L636` in this same branch the sentinel is read. BK models embed their display lists inline as raw N64 command words (the `mRawDLWords` blob that the importer widens at load time), so a texture reference is just a segmented address buried in a `G_SETTIMG` word. There's no factory pass to swap opcodes into, so I put the marker in the one field that survives the raw blob (w1's high byte) and decoded it in BK's own importer. Torch doesn't export models the way other ports do because the BK decomp expects this format.

Dynamically-textured DLs like the falling jiggies transition are left un-stamped on purpose. They get their texture at runtime, so stamping them effectively disables that feature.

src/port/UI/LighthouseMenuSettings.cpp
```diff
@@ -478,25 +479,35 @@ void LighthouseMenu::AddMenuSettings() {
         })
         .Options(ButtonOptions().Tooltip("Displays a test notification."));
 
-    // Mod Menu
-    path.sidebarName = "Mod Menu";
+    // Romhack Menu
+    path.sidebarName = "Romhack Menu";
```
Malkierian, line 483: This rename doesn't make sense, it's still a mod menu. There will be mods other than romhacks.

src/port/UI/LighthouseMenuSettings.cpp
```diff
@@ -478,25 +479,35 @@ void LighthouseMenu::AddMenuSettings() {
...
     AddSidebarEntry("Settings", path.sidebarName, 1);
 
-    AddWidget(path, "Generate Mod from ROM", WIDGET_BUTTON)
+    AddWidget(path, "Generate Romhack from ROM", WIDGET_BUTTON)
```
Malkierian: I didn't see this until testing other things, but this should be part of the MenuInit in the ModMenuWindow, so that it's after the window popout button and included in the popout should it happen.

src/port/UI/LighthouseModMenuWindow.cpp
```diff
@@ -478,13 +623,47 @@ void LighthouseModMenuWindow::DrawElement() {
...
+}
+
 void LighthouseModMenuWindow::InitElement() {
     UpdateModFiles(true);
 }
 
 static void RegisterModMenuWidgets() {
     enableModsWidget = { .name = "Enable Mods", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
-    enableModsWidget.CVar(CVAR_SETTING("AltAssets"))
+    enableModsWidget.CVar("gEnhancements.Mods.AlternateAssets")
```
Malkierian: 1) Why change this to enhancements if it's in settings?
2) Use the macro, please XD.

src/port/UI/LighthouseMenuSettings.cpp
JeodC: It's not a rename, the Mod Menu is further down. Git just reported the change that way.

PR #167, opened 2026-06-17
Add control schemes (Issue #152)

### LINKED ISSUE #152: [ENHANCEMENT] Combobox for control schemes: Retro, Xbox, Pocket
Add an enhancement/general controls combobox dropdown for the following control schemes:

| Retro | Xbox | Pocket | Function |
|-------|-------|---------|-----------|
| Left stick | Left stick | D-Pad | Move, swim, climb |
| C-UP | Y | Y | First person view |
| C-Left/Right | Right stick | L1/R1 | Move camera left/right |
| C-Down | Right stick | Select | Zoom camera in/out |
| R | LB/RB | B | Center camera behind Banjo |
| Z | LT/RT | L2 | Hold to crouch, change direction with Move, press other trigger to activate Talon Trot |
| A | A | A | Jump, swim, fly, shock jump |
| B | X | X | Claw swipe, dive, cancel |

JeodC:
Testing of all control schemes verified all button actions retained and working well. Resolve #152

### DIFF
--- a/src/core1/io/pfsmanager.c
+++ b/src/core1/io/pfsmanager.c
@@ -8,7 +8,6 @@
 #include "version.h"
 
 #include "port/Patches/Patches.h"
-#include "port/Enhancements/Camera/FreeLookCamera.h"
 
 #define PFSMANAGER_THREAD_STACK_SIZE 0x200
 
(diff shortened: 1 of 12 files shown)


Malkierian:
So how does this actually apply? Do you select the combobox then hit Reset Controls?

JeodC:
> So how does this actually apply? Do you select the combobox then hit Reset Controls?

Just select the option, no hitting any apply button.

Malkierian:
I not so sure that would be a good idea from the UX perspective. It changes a lot in controls, and to have that not require a confirmation to apply seems dangerous.

JeodC:
I can add an confirmation, but I don't think it's necessary. It takes some effort to click the dropdown and then click on a scheme.

Malkierian:
Some kind of apply button for comboboxes that change other things is also standard practice for the rest of the projects, so it would also make for better consistency.

PR #182, opened 2026-06-19
Final boss dev sequence and fix enhancement

JeodC:
https://discord.com/channels/808039310850130000/1515867977524838451/1517519490940801135

Add dev tools warp buttons to finale phases and a fix enhancement for the "Jinjo Statue Softlock" bug, well known in the speedrun community.


src/port/DevTools/Sequences.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+// Registration
+
+void RegisterSequences_Init() {
+    REGISTER_LISTENER(GameFrameUpdate, EVENT_PRIORITY_NORMAL, [](IEvent*) {
+        TickSequences();
+        TickFinalBoss();
+    });
```
Malkierian, line 277: GameFrameUpdate definitely should not have ungated listeners in almost all cases.

JeodC: Does the change align more with what you're looking for?

Malkierian: No, they're still virtually unconditional, you just made sure they wouldn't get registered more than once. The `sHooksInstalled` paradigm isn't necessary with COND_HOOK, it handles that with its own internal variables. The issue is that this hook is registered every time the init is called, but I *know* that the functionality that these are enabling is conditional in dev tool stuff.

JeodC:
Draft, unsure if this is needed any longer. Relevant: https://discord.com/channels/808039310850130000/1007412519914197012/1518041337084121099

Malkierian:
Probably better just to paste that post in here, since that's a private channel.

JeodC:
> Probably better just to paste that post in here, since that's a private channel.

Fair.

This is parked until the event system gains a safe deferred-unregister path (or the warp buttons accept staying registered as a permanent early-return), and pending a decision on whether the buttons are worth keeping at all. The feature adds dev-tool warp buttons for the final boss fight. Installing the hook is easy (register it on the button press), but the listener can't cleanly remove itself when it's done. It discovers it's finished from inside its own tick i.e. mid-dispatch, while the event system is looping over the listener list. Calling `UNREGISTER_LISTENER` at that point mutates the list currently being iterated, causing a concurrency crash. The usual `COND_HOOK/COND_VB_SHOULD` pattern doesn't help here because it only unregisters when a CVar-driven init re-runs, not in response to runtime state discovered during a tick.

PR #198, opened 2026-06-23
[FEATURE] Implement Randomizer

Caladius:
This adds a wealth of events and options specific to Rando functionality.

**New Shuffles**
- Empty Honeycombs
- Jiggies
- Jinjos
- Mumbo Tokens
- Music Notes
- Molehill Abilities 

**Logic Types**
- Glitchless
-  No Logic

Rando relies on its own retention for Jinjos and Music Notes, it allows for cross world collection and will recalculate notes upon world entry and will spawn a Jinjo Jiggy for any world when 5 Jinjos for a given world has been collected.

Molehills are shuffled amongst themselves when enabled, this will randomize which ability you learn from each.


src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+
+    return NULL;
+}
+
+Actor* FindActorByRandoCheckId(RandoCheckId randoCheckId) {
+    Actor* start;
+    Actor* end;
+    Actor* baddieActor;
```
JeodC: I couldn't find where `baddieActor` is initialized. If `CheckSpawnedIdList()` is false, the loop is skipped and L139 reads an uninitialized pointer. Also, when the loop runs but finds nothing it exits as `end` (not null), so the `== NULL` fallback spawn doesn't fire.

Should this initialize to `NULL` and track "found" separately?

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
```diff
@@ -0,0 +1,537 @@
...
+    REGISTER_LISTENER(OnFindActorMarkerFromJiggyId, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        OnFindActorMarkerFromJiggyId* ev = (OnFindActorMarkerFromJiggyId*)event;
+
+        if (!IS_RANDO) {
+            return;
+        }
+
+        switch (ev->jiggyId) {
+            case JIGGY_3E_GV_GRABBA:
+                ev->result = FindActorByRandoCheckId(RC_GV_JIGGY_GRABBA)->marker;
```
JeodC: This looks like it dereferences a function that can return `NULL` (L155). Add a check like the siblings at L411 and L493?

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
```diff
@@ -0,0 +1,537 @@
...
+        Rando::StaticData::RandoStaticCheck randoStaticCheck = Rando::StaticData::Checks[randoCheckId];
+        RandoSaveCheck randoSaveCheck = RANDO_SAVE_CHECKS[randoCheckId];
+        actor_e actorId = (actor_e)Rando::StaticData::Items[randoSaveCheck.randoItemId].actorId;
+
+        int32_t position[3];
+        position[0] = randoStaticCheck.posX;
+        position[1] = randoStaticCheck.posY + 50;
+        position[2] = randoStaticCheck.posZ;
+
+        Actor* newActor = CustomObject::SpawnCustomActorEX(randoCheckId, position, &actorInfoMap.at(actorId).first,
```
JeodC, line 139: `actorInfoMap.at(actorId)` throws when the check's `randoItemId` resolves to `ACTOR_1` or `ACTOR_12C` (not in the map). This would be reachable with an unshuffled check or a corrupt/edited save. To guard, use `find()` instead of `at()` (same pattern at Bundle.cpp:286).

src/port/Rando/Logic/GeneratePools.cpp
```diff
@@ -0,0 +1,170 @@
...
+        Rando::Logic::ShuffleRandoItems(CVarGetString(CVAR_RANDOMIZER_SETTING("InputSeed"), ""), abilityItemPool);
+    }
+
+    if (RANDO_SAVE_OPTIONS[RO_LOGIC].optionValue == RO_LOGIC_GLITCHLESS) {
+        Rando::Logic::GenerateGlitchlessLogicPool(checkPool, itemPool, abilityCheckPool, abilityItemPool, saveData);
+    } else if (RANDO_SAVE_OPTIONS[RO_LOGIC].optionValue == RO_LOGIC_NO_LOGIC) {
+        Rando::Logic::GenerateNoLogicPool(itemPool, abilityItemPool);
+    }
+
+    for (int i = 0; i < checkPool.size(); i++) {
```
JeodC, line 106: This loop looks to be bounded by `checkPool.size()` but it indexes `itemPool[i]`. On a failed glitchless seed, the failure branch never rebuilds `itemPool`, so this reads out of bounds.

src/port/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -0,0 +1,665 @@
...
+                            a = 0;
+                            failSafeTrigger = false;
+                        }
+                    }
+                }
+            } else {
+                if (prevProgressionIndex == progressionIndex) {
+                    Notification::Emit({ .message = "Seed Configuration impossible, failed to generate." });
+                    RefreshMetrics("Seed Failed to Generate");
+                    ResetSaveData();
```
JeodC, line 653: This impossible seed path resets save data without restoring `itemPool` to `checkPool` size, which will make the `GeneratePools` loop at `GeneratePools.cpp:102` read OOB.

src/core2/ch/mole.c
```diff
@@ -471,14 +473,16 @@ void chmole_update(Actor *this){
 
 int chmole_learnedAllSpiralMountainAbilities(void){
```
JeodC, line 474: This needs to return true after the `if` block to prevent falling off the end of a value-returning function. This wasn't needed previously because the function would always return something without the `if` conditional gate.

src/port/Rando/Rando.h
```diff
@@ -0,0 +1,35 @@
...
+extern int16_t selectedFileNum;
+
+#define DEFAULT_FILE_NUM -1
+
+#define IS_RANDO                         \
+    (selectedFileNum == DEFAULT_FILE_NUM \
+         ? false                         \
+         : gameFile_saveData[selectedFileNum].shipSaveData.fileType == FILE_TYPE_SAVE_RANDO)
+
+#define RANDO_SAVE_CHECKS gameFile_saveData[selectedFileNum].shipSaveData.randoSaveData.randoSaveCheck
```
JeodC, line 21: If this is ever used at file select it can hit OOB since there is no `-1` guard. `IS_RANDO` uses that, maybe it should be used here for futureproofing.

src/port/Rando/Logic/Metrics.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    learnedAbilities.clear();
+    for (int a = ABILITY_0_BARGE; a < ABILITY_13_1ST_NOTEDOOR; a++) {
+        if (ability_isUnlocked((ability_e)a)) {
+            learnedAbilities.push_back(abilityNameList[a].c_str());
+        }
+    }
+
+    regionAccessList.clear();
+    for (int r = RR_UNKNOWN; r < RR_MAX; r++) {
```
JeodC, line 135: The loop here starts at `RR_UNKNOWN`, whose `Regions[]` entry has a `nullptr` name. If `std::string(nullptr)` is ever pushed back, this would crash. Start at +1 instead if possible.

src/core2/actor_cubepropsystem.c
```diff
@@ -475,7 +479,14 @@ Prop *__codeA5BC0_initProp2Ptr(Cube *cube) {
         cube->prop2Cnt = 1;
         cube->prop2Ptr = bk_malloc(sizeof(Prop));
     }
-    sp1C = &cube->prop2Ptr[cube->prop2Cnt-1];
+    // port - Rando manipulation of Props leads to rare errors with prop2Ptr, this ensures prop2Ptr is allocated each time if NULL
+    if (cube->prop2Ptr == NULL) {
+        cube->prop2Ptr = bk_malloc(sizeof(Prop));
```
JeodC, line 484: The guard here masks a `prop2Cnt` / `prop2Ptr` desync rather than fixing it. If `prop2Cnt > 1` while the pointer was `NULL`, then later access is OOB. The cube claims N props but the buffer only holds one, so anything that walks `prop2Ptr[0..prop2Cnt-1]` (the removal path, the per-frame loops) reads/writes off the end and smashes the heap. I think it's better to reset `prop2Cnt` to match what was actually allocated, and ideally track down whatever desyncs the two during rando prop manipulation in the first place.

src/core2/level/specificflags.c
```diff
@@ -77,7 +77,9 @@ void _levelSpecificFlags_updateCRC2(void) {
 }
 
 s32 levelSpecificFlags_get(s32 i){
-    return bitfield_get_bit(D_80383320.unk8, i);
+    CALL_CANCELLABLE_RETURN_EVENT(OnGetLevelSpecificFlag, i) {
+        return bitfield_get_bit(D_80383320.unk8, i);
+    }
```
Malkierian, line 82: 1) Does this return anything if it is cancelled?
2) Why would you want to cancel the return of this? Shouldn't any special handling happen on the other end of an event?

src/core2/map/cutscene_skip.c
```diff
@@ -58,6 +63,10 @@ bool cutscene_skipGameOverCutsceneCheck(void) {
     if (mapSpecificFlags_get(0) != 0) {
         fileProgressFlag_set(FILEPROG_E1_UNKNOWN, 1);
     }
+    bool skipMiscCutscenes = false;
+    CALL_EVENT(OnMiscCutscenesCheck, &skipMiscCutscenes);
+    if (skipMiscCutscenes && func_8024E698(0) == 1) { return true; }
+
```
Malkierian, line 69: These two should probably be VBs in the future.

src/core2/map/warp_dispatch.c
```diff
@@ -561,6 +561,7 @@ void func_80334448(NodeProp *arg0, ActorMarker *arg1) {
                 // [port] BB romhacks remap warp destinations via BKCF config
                 warpIdx = codeA5BC0_getNodePropUnk8(arg0);
                 warpDest = port_getRomhackWarpDest(warpIdx);
+                CALL_EVENT(OnWarpDispatch, warpIdx, warpDest);
```
Malkierian, line 564: Doesn't have to happen now, but this might need to be cancellable in the future.

src/core2/sprite/rendertex.c
```diff
@@ -74,8 +74,10 @@ s32 spriteRender1PrimMode;
 void func_80347DF0(Struct81s *arg0){
     Actor *actor;
     actor = bundle_spawn_f32(arg0->actor_id, arg0->position);
-    actor->unk54 = 3.0f;
-    func_803333DC(arg0, actor);
+    if (!EventSystem_Should(VB_BUNDLE_SPAWN_SET_ACTOR_DATA, true, actor)) {
+        actor->unk54 = 3.0f;
+        func_803333DC(arg0, actor);
+    }
```
Malkierian, line 80: May be functional now, but this seems backwards. The name and return of the VB should reflect the vanilla flow. I have trouble understanding what's happening with this ! at the start.

src/core2/level/specificflags.c
Caladius: This is handled in our WorldState.cpp file. Currently there is a single Jiggy in the Christmas Tree in Freezeezy Peak that has no collision enabled if you havent completed the Tree minigames. This collision toggle affect that Jiggy wherever it is in the world so we override it to only do that for the item in the vanilla location and default the flag to ON if its anywhere else.

Otherwise it lets the function do what it does normally.

src/core2/map/warp_dispatch.c
Caladius: Agreed, this was a placeholder for Entrance Shuffle but we ended up using it for something now that just needs to know when this happens.

src/core2/sprite/rendertex.c
Caladius: I think this can be inverted now to match your expectation. I was very confused with this early on and I think i figured out what the underlying issue was. Added to my notes to address.

src/port/Rando/Logic/Metrics.cpp
```diff
@@ -0,0 +1,155 @@
...
+std::string statusText = "";
+int32_t checksInPool = 0;
+int32_t itemsInPool = 0;
+int32_t abilitiesInPool = 0;
+
+std::vector<std::string> learnedAbilities;
+std::vector<std::string> regionAccessList;
+std::vector<std::string> eventAccessList;
+
+#define WIDGET_TEXT_COLOR(color) UIWidgets::ColorValues.at(UIWidgets::Colors::##color)
```
JeodC: Compile error: `error: pasting formed '::Green', an invalid preprocessing token`

src/port/Rando/Logic/Metrics.cpp
```diff
@@ -0,0 +1,155 @@
...
+            ImGui::Text(ability.c_str());
+        }
+        ImGui::EndChild();
+    }
+}
+
+void Metrics_DrawSeedData() {
+    if (ImGui::BeginChild("SeedData", ImVec2(ImGui::GetContentRegionAvail().x, 0))) {
+        ImVec4 statusColor = seedStatus         ? WIDGET_TEXT_COLOR(Green)
+                             : checksInPool < 0 ? WIDGET_TEXT_COLOR(Red)
```
JeodC, line 51: Clang compile: `error: pasting formed '::Red', an invalid preprocessing token`

src/port/Rando/Logic/Metrics.cpp
```diff
@@ -0,0 +1,155 @@
...
+                                                : WIDGET_TEXT_COLOR(Orange);
```
JeodC, line 52: Clang compile: `error: pasting formed '::Orange', an invalid preprocessing token`

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
```diff
@@ -0,0 +1,537 @@
...
+bool nextActorSaveState = false;
+
+void LogOutSpawns(int32_t actorId, int16_t posX, int16_t posY, int16_t posZ) {
+    std::string locationStr = std::to_string(posX) + ", " + std::to_string(posY) + ", " + std::to_string(posZ);
+    BK_LOG_INFO("Actor ID: %i | Position: %s", actorId, locationStr.c_str());
+}
+
+void LogOutCollision(int32_t actorId, int16_t posX, int16_t posY, int16_t posZ) {
+    std::string locationStr = std::to_string(posX) + ", " + std::to_string(posY) + ", " + std::to_string(posZ);
+    BK_LOG_INFO("Collect ID: %i | Position: %s", actorId, locationStr);
```
JeodC: Clang compile: `error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]`

src/port/Rando/StaticData/Flags.cpp
```diff
@@ -0,0 +1,35 @@
...
+    RF(RANDO_INF_ANCHOR_RAISED,             RO_GENERIC_OFF),
+    RF(RANDO_INF_CLANKER_RAISED,            RO_GENERIC_OFF),
+    RF(RANDO_INF_MINIGAME_RINGS_COMPLETED,  RO_GENERIC_OFF),
+    RF(RANDO_INF_WATER_PYRAMID_DRAINED,     RO_GENERIC_OFF),
+};
+// clang-format on
+
+RandoInf GetFlagIdFromName(const char* name) {
+    for (auto& [randoInf, randoStaticFlag] : Flags) {
+        if (strcmp(name, randoStaticFlag.name) == 0) {
```
JeodC: Clang compile: `error: use of undeclared identifier 'strcmp'`

src/port/Rando/StaticData/Options.cpp
```diff
@@ -0,0 +1,38 @@
...
+    RO(RO_SHUFFLE_JINJOS,           RO_GENERIC_OFF),
+    RO(RO_SHUFFLE_MOLEHILLS,        RO_GENERIC_OFF),
+    RO(RO_SHUFFLE_MUMBO_TOKENS,     RO_GENERIC_OFF),
+    RO(RO_SHUFFLE_MUSIC_NOTES,      RO_GENERIC_OFF),
+};
+// clang-format on
+
+RandoOptionId GetOptionIdFromName(const char* name) {
+    for (auto& [randoOptionId, randoStaticOption] : Options) {
+        if (strcmp(name, randoStaticOption.name) == 0) {
```
JeodC: Clang compile: `error: use of undeclared identifier 'strcmp'`

src/port/Rando/LighthouseMenuRando.cpp
```diff
@@ -0,0 +1,264 @@
...
+            }
+            Ship::Context::GetRawInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
+        }
+        ImGui::PopID();
+
+        if (ImGui::BeginTable("Abilities Table", 3, ImGuiTableFlags_SizingStretchSame)) {
+            ImGui::TableNextColumn();
+
+            for (auto& [abilityId, abilityInfo] : abilityLoadoutMap) {
+                abilityToolTip = std::format("Start with {} unlocked.", abilityInfo.first);
```
JeodC: Clang compile: `error: no member named 'format' in namespace 'std'`

src/port/Rando/LighthouseMenuRando.cpp
```diff
@@ -0,0 +1,264 @@
...
+            }
+            Ship::Context::GetRawInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
+        }
+        ImGui::PopID();
+
+        if (ImGui::BeginTable("Items Table", 3, ImGuiTableFlags_SizingStretchSame)) {
+            ImGui::TableNextColumn();
+
+            for (auto& [itemId, itemInfo] : itemLoadoutMap) {
+                itemToolTip = std::format("Start with max {}.", itemInfo.first);
```
JeodC: Clang compile: `error: no member named 'format' in namespace 'std'`

JeodC (changes requested):
Did not include the case sensitive includes.

src/core2/savedata.c
```diff
@@ -438,9 +439,11 @@ int savedata_8033CE40(void *buffer){
 }
 
 void savedata_clear(void *savedata_){
-    u8 *savedata = (u8 *)savedata_;
-    int i;
-    for(i = 0; i < sizeof(SaveData); i++){
-        savedata[i] = 0;
+    CALL_CANCELLABLE_RETURN_EVENT(OnSaveClear, savedata_) {
```
Malkierian: Shouldn't this just be a cancellable event rather than a return event?

src/core2/bundle.c
```diff
@@ -144,39 +144,42 @@ Actor *__bundle_spawnWithFirstActor(enum bundle_e bundle_id, f32 position[3], Ac
...
-        actor->yaw = bundle->yaw = (bundle_info->flags & 0x20) ? bundle_info->yaw : randf2(0.0f, 360.0f);
-        bundle->elapsed_time = 0.0f;
-        bundle->unk2C = 0;
-        bundle->unk2D = 1;
-        bundle->flags = bundle_info->flags;
-        bundle->unk2E = (bundle_info->flags & 0x1) ? (0.5 < randf()) : 0;
-
-        if (bundle_info->flags & 0x200) {
-            actor->unk5C = bundle->position[1];
+        if (!EventSystem_Should(VB_OVERRIDE_BUNDLE_SPAWN, false, bundle_id, bundle_info, i, position, &actor)) {
```
Malkierian, line 147: Another seeming inversion.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: if CheckSpawnedIdList() is false then baddieActor is NULL and L139 forces an actor to Spawn to be assigned into that baddieActor.

src/core2/ch/mole.c
```diff
@@ -219,27 +219,29 @@ void chmole_additionalAbilityLearnActions(ActorMarker *marker, enum asset_e arg1
...
-        sp2C = moleTable[this->actorTypeSpecificField-9].refresher_text_id;
-    }//L802D99EC
-    // New Ability: Learn Dialog & Misc Actions
-    else{
-        func_80347A14(0);
-        this->has_met_before = true;
-        sp2C = moleTable[this->actorTypeSpecificField-9].teach_text_id;
-        ability_unlock(moleTable[this->actorTypeSpecificField-9].ability);
-        switch(moleTable[this->actorTypeSpecificField-9].ability){
+    if (!EventSystem_Should(VB_OVERRIDE_MOLEHILL_ABILITY, false, this, &sp2C, &sp28)) {
```
Malkierian, line 222: And another inversion.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: If this returns NULL the event doesnt cancel and the vanilla code takes over, in Rando this shouldnt ever be an issue.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: At this point in the flow we've already safe guarded an unknown RandoCheckId from being passed through.

src/port/Rando/Logic/GeneratePools.cpp
Caladius: Do you have a seed that failed to generate? In this flow when one fails the game loads vanilla, you need to reset and return to the File Select and try again which empties out all the lists and tries again fresh.

src/port/Rando/Logic/GlitchlessLogic.cpp
Caladius: Oversight, should have added a matching check to ensure checkPool and itemPool sizes match. Good call out.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: These were both created for testing, can probably be removed.

src/port/Rando/StaticData/Flags.cpp
Caladius: I dont think Im using these anyway, ill remove em

src/port/Rando/StaticData/Options.cpp
Caladius: removed

src/port/Rando/LighthouseMenuRando.cpp
Caladius: Does this not exist in Linux? It works on Windows so I'm not sure (genuine question)

src/port/Rando/LighthouseMenuRando.cpp
Caladius: same her:
Does this not exist in Linux? It works on Windows so I'm not sure (genuine question)

src/core2/savedata.c
Caladius: I think we want to preserve the SaveData to be passed onto the next step in the vanilla flow. Honestly I don't know, mainly I didnt want to just kill it haha.

src/core2/bundle.c
Caladius: This is written that same way as the VB_UPDATE_JINJO_HUD is that you changed haha.

src/core2/ch/mole.c
Caladius: This is written that same way as the VB_UPDATE_JINJO_HUD is that you changed haha.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
JeodC: I don't think so. A local `Actor* baddieActor;` is not zero-initialized in C/C++, it holds an indeterminate value. When `CheckSpawnedIdList()` is false the loop never runs, so `baddieActor` is never assigned, and `if (baddieActor == NULL)` at L139 reads garbage. It's not guaranteed NULL. Plus, when the loop *does* run, it exits with `baddieActor == end` (which isn't NULL).

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
JeodC: But the crash precedes the cancel check. `ev->result = FindActorByRandoCheckId(...)->marker;` dereferences `->marker` on the return value before the `if (ev->result != NULL)` guard runs.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
JeodC: My bad, I mean the throw is from the *item* at that check resolving to one of the two actors. Granted, those only happen when the check was never shuffled, but using `find()` over `.at()` is still a nice insurance policy.

src/port/Rando/Logic/GeneratePools.cpp
JeodC: Looks to me that the reset happens after the OOB has already executed.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: Fixed in pending changes.

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: fixed in pending changes

src/port/Rando/ObjectBehavior/ObjectBehavior.cpp
Caladius: I don't think this is an issue given all the guards before it gets here. If you stumble across something that trips it let me know and I'll address it specifically.

src/core2/bundle.c
Caladius: TODO: Look at later down the road

src/core2/ch/mole.c
Caladius: TODO: Look at later down the road

src/port/Rando/MiscBehavior/OnFileSave.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+void Rando::MiscBehavior::OnFileSave() {
+    REGISTER_LISTENER(OnGameSave, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        OnGameSave* ev = (OnGameSave*)event;
+
+        if (!IS_RANDO) {
+            return;
+        }
+    });
+}
```
Malkierian, line 16: Is this even needed anymore? Are you expecting to need to use it in the future?

src/core2/actor_cubepropsystem.c
Caladius: In Rando's flow this is necessary as we will most likely manipulate its data post creation. Unfortunately going down the hole for this issue would more than likely require more source intrusion that this is doing.

src/port/Rando/Logic/Metrics.cpp
Caladius: RR_UNKNOWN is skipped for all calculations and is just used a "if randoRegionId == RR_UNKNOWN then continue" purposes. Should be safe.

src/port/Rando/Rando.h
Caladius: For this we safeguard it but forcibly locking these out if selectedFileNum is -1 which occurs upon load of the FileSelect screen.

src/core2/ch/mole.c
Caladius: The Hook here returns a value when in Rando, otherwise it follows its source vanilla code flow. Shouldnt need any additional guards here.

src/port/Rando/MiscBehavior/OnFileSave.cpp
Caladius: Its currently not in use. I do plan on using it in the future but I can nuke it if you'd rather us add it back when needed.

edit: This is a placeholder for Spoiler Log actions.

src/port/Rando/LighthouseMenuRando.cpp
```diff
@@ -0,0 +1,266 @@
...
+const char* logicModes[2] = {
+    "Glitchless",
+    "No Logic",
+};
+
+std::map<int32_t, const char*> logicModeMap = {
+    { RO_LOGIC_GLITCHLESS, "Glitchless" },
+    { RO_LOGIC_NO_LOGIC, "No Logic" },
+};
+std::map<const char*, int32_t> logicModeMapa = {
```
Malkierian: This doesn't seem to be used, and I can't think of a reason why it would.

PR #247, opened 2026-07-09
Decomp pass 5

JeodC:
Compile tested in Windows


JeodC:
I'm labeling this for the 1.0 milestone because it realigns our decomp codebase with upstream decomp with filenames and paths. Previously we diverged because we wanted to find code more easily with friendly filenames. Now that release is near, we should stick with upstream decomp so that the general public has familiarity.

PR #256, opened 2026-07-19
Add a differ in Default and Authentic boot logos

JeodC:
In Default, introduce Litwise to replace the N Logo actor in intro, and replace Mumbo's xylophone texture.


src/core2/fx/cutscene_actors.c
```diff
@@ -121,7 +121,13 @@ void func_802E07E0(Actor *this) {
...
             func_802C7318(this);
         }
-        func_80361E10(this);
+        // [port] Let an enhancement suppress this decor actor's animation-driven sound sequence.
+        if (EventSystem_Should(VB_INIT_ANIM_SOUND_SEQUENCE, true, this->marker->modelId)) {
+            func_80361E10(this);
+        } else {
+            this->unk134 = NULL;
+            this->unk160 = NULL;
+        }
```
Malkierian: Why not set this up to pass the Actor pointer instead, so the else block can be handled in the VB handler?

src/core2/actor_cubepropsystem.c
```diff
@@ -1688,7 +1688,12 @@ BKModelBin *marker_loadModelBin(ActorMarker *this){
     // [port] Let the language system invalidate a re-pointed model so it
     // re-fetches the active language's version (live swap on the next draw).
     s32 reload = 0;
-    CALL_EVENT(OnModelLoad, this->modelId, modelInfo, &reload);
+    s32 skip = 0;
+    CALL_EVENT(OnModelLoad, this->modelId, modelInfo, &reload, &skip);
+    // [port] A listener can suppress a model entirely
+    if (skip) {
+        return NULL;
+    }
```
Malkierian: Maybe a cancellable instead? Doesn't have to be, just thought I'd bring it up.

PR #273, opened 2026-07-26
Add support for imgui file picker fallback

JeodC:


JeodC:
Will be rolling this into a separate bundled PR with a LUS bump of its own.

PR #274, opened 2026-07-27
Rc fixes 2

JeodC:


src/port/Engine.cpp
```diff
@@ -746,13 +753,54 @@ void GameEngine::RunExtract(int argc, char* argv[]) {
                         continue;
                     }
                     case PS_FIRST: {
-                        if (args.empty() && !extract.SelectGameFromUI()) {
-                            promptStep = PS_FILE_CHECK;
+                        if (args.empty()) {
+                            // Skip the picker entirely if a baserom.us.z64 is sitting in the app
+                            // directory: load it and go straight to extraction.
```
Malkierian, line 758: Why limit it to that specific filename? Does the root directory ROM search not already work for this, with the *.z64 checks?

JeodC: Lighthouse was already set up to look for the specific name. I just mirrored it.

Malkierian: I suppose that'll be fine for now, but the new menu flow should have had the ability to search for ROM files to attempt to process like the Zeldas do, since that's where it was originally developed.

PR #277, opened 2026-07-28
Apply modern camera changes back to free look.

Malkierian:


JeodC:
Does this still fight with modern camera when both are enabled?

Malkierian:
Doesn't look like it. Pitch doesn't change with free look enabled on modern controls but camera is moving in and out.

Malkierian:
Was incidental, though. I wasn't *trying* to fix that part of it.

Malkierian:
nevermind, I got it to do both after all.

PR #292, opened 2026-07-30
Extend prop count

Malkierian:
Extends propCnt1/propCnt2 to prevent overflow from particle creation that results in stale references trying to be freed.

### DIFF
--- a/include/prop.h
+++ b/include/prop.h
@@ -406,11 +406,12 @@ typedef struct cude_s{
     s32 x:5;
     s32 y:5;
     s32 z:5;
-    // [port] were :6 bitfields; the 64th prop wrapped the count to 0, which
-    // realloc'd the prop array to nothing while live markers still pointed into it.
+    // [port] were bitfields (:5/:6/:6). The 64th prop wrapped a count to 0, which
+    // realloc'd the prop array to nothing while live markers still pointed into it;
+    // unk0_4 is the front index of the node-prop partition and wrapped at 32.
+    u16 unk0_4; //node_prop_count
     u16 prop1Cnt;
     u16 prop2Cnt;
-    u32 unk0_4:5; //node_prop_count
     NodeProp *prop1Ptr;
     Prop *prop2Ptr;
 }Cube;
--- a/include/prop.h
+++ b/include/prop.h
@@ -409,7 +409,7 @@ typedef struct cude_s{
     // [port] were bitfields (:5/:6/:6). The 64th prop wrapped a count to 0, which
     // realloc'd the prop array to nothing while live markers still pointed into it;
     // unk0_4 is the front index of the node-prop partition and wrapped at 32.
-    u16 unk0_4; //node_prop_count
+    u32 unk0_4:16; //node_prop_count
     u16 prop1Cnt;
     u16 prop2Cnt;
     NodeProp *prop1Ptr;
--- a/include/prop.h
+++ b/include/prop.h
@@ -398,12 +398,18 @@ typedef struct {
     u32 pad10_0:1; 
 } OtherNode; //can be inplace of NodeProp (see code7AF80_initCubeFromFile) size: 12 (0xC) bytes
 
+// Distance scratch __cube_sort fills, one entry per prop. Cubes past this are
+// left unsorted rather than overrunning it.
+#define CUBE_SORT_SCRATCH_SIZE 0x400
+
 typedef struct cude_s{
     s32 x:5;
     s32 y:5;
     s32 z:5;
-    u32 prop1Cnt:6;
-    u32 prop2Cnt:6;
+    // [port] were :6 bitfields; the 64th prop wrapped the count to 0, which
+    // realloc'd the prop array to nothing while live markers still pointed into it.
+    u16 prop1Cnt;
+    u16 prop2Cnt;
     u32 unk0_4:5; //node_prop_count
     NodeProp *prop1Ptr;
     Prop *prop2Ptr;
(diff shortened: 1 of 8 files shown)


src/core2/actor_cubepropsystem.c
```diff
@@ -524,6 +529,11 @@ s32 func_8032D9C0(Cube *cube, Prop* prop){
         // port - Rando manipulation of prop2Ptr results in mismatched data, this corrects that.
         ptrdiff_t index = prop - cube->prop2Ptr;
         ptrdiff_t elementsAfter = (cube->prop2Cnt - 1) - index;
+        // [port] A stale propPtr makes elementsAfter enormous; drop the removal instead.
+        if (index < 0 || index >= (ptrdiff_t)cube->prop2Cnt) {
```
JeodC, line 526: This guard fires too late. By the time we get here, prop has already been dereferenced twice on lines 524 and 526, and `func_80305CD8(..., -1)` has run. On the exact path this guard exists to catch, we've read freed memory and decremented `sProp2TotalCounts[idx]` with an idx derived from `prop->unk4[1]` out of that same stale pointer, i.e. an arbitrary band. The matching +1 is in `func_8032F21C`, so the early return leaves that counter permanently under. A NULL propPtr also crashes on 524 before ever reaching here.

Move the check to the top of the `if(cube->prop2Cnt != 0)` block, before any use of prop, and fold in `prop == NULL`. `marker_free` passes `this->cubePtr` unchecked too, so `cube == NULL` costs nothing to add while you're here.

include/prop.h
```diff
@@ -398,12 +398,18 @@ typedef struct {
...
+
 typedef struct cude_s{
     s32 x:5;
     s32 y:5;
     s32 z:5;
-    u32 prop1Cnt:6;
-    u32 prop2Cnt:6;
+    // [port] were :6 bitfields; the 64th prop wrapped the count to 0, which
+    // realloc'd the prop array to nothing while live markers still pointed into it.
+    u16 prop1Cnt;
```
JeodC, line 413: A non-bitfield `u16` can't share the storage unit with `x/y/z`, so it opens a new one and `unk0_4:5` opens a third. `sizeof(Cube)` goes from 24 to 32 on 64-bit. Moving `unk0_4` above the counts puts it back at 24 with no other change:

```
s32 x:5;
s32 y:5;
s32 z:5;
u32 unk0_4:5; //node_prop_count
u16 prop1Cnt;
u16 prop2Cnt;
```
This is worth doing because `sCubeList.cubes` is one `bk_malloc` of `cubeCnt * sizeof(Cube)` out of the fixed HEAP_SIZE 0x210520 (~2.1 MB) heap in `src/core1/memory.c`. A 9,000-cube map pays +72 KB, ~3.5% of the whole heap, on a budget we already had to stretch for the rando marker pool.

include/prop.h
```diff
@@ -398,12 +398,18 @@ typedef struct {
...
+    u16 prop2Cnt;
     u32 unk0_4:5; //node_prop_count
```
JeodC: `unk0_4` is still `:5` while `__codeA5BC0_initPropPointerForCube` increments it once per non-special node prop, up to cnt (a u8). This PR newly exposes it: under the old `:6`, a cube with ≥64 node props made `code7AF80_initCubeFromFile` read `cube->prop1Cnt * sizeof(NodeProp) = 0` bytes and desync the file stream, so no shipped map has one. Now that prop1Cnt is u16 those cubes load correctly and `unk0_4` can be pushed past 31 and wrap. It's the same class of bug you're fixing here, and worth widening in the same pass.

src/port/Patches/PerfPatches.cpp
```diff
@@ -27,6 +30,15 @@ static f32 sPadFrustumY = 93.9692611694336f;
 static f32 sPadPlanes[4][4];
 static bool sPadPlanesDirty = true;
 
+// Capped so a bad frame can't flood the log; one print is enough to investigate.
+extern "C" void port_warnPropNotInCube(int32_t index, int32_t propCnt) {
```
JeodC, line 35: Could this find a better home? This isn't a performance related patch. The rest of the file is frustum culling, lighting, and sort.

JeodC (changes requested):

src/core2/actor_cubepropsystem.c
Malkierian: I can't move that check up because it depends on the index variable that is defined and initialized just two lines up, but I can put that null check next to the prop2Cnt check.

src/port/Patches/PerfPatches.cpp
Malkierian: I mean, it *kind of* fits, because this came about for changing the max number of particles that can be spawned at once.

src/core2/actor_cubepropsystem.c
JeodC: `index` only needs `prop` and `cube->prop2Ptr`, both of which you have at function entry. You can move the computation up with the check, not just the check. The null check next to `prop2Cnt` does fix the crash case, but the stale-pointer case is unchanged: `prop->unk8_1` and `prop->unk4[1]` are still read out of freed memory on 524/526, and `func_80305CD8(..., -1)` still decrements `sProp2TotalCounts` for whatever band that garbage resolves to, before the guard on 533 ever runs. The return also hands back sp24 read from the stale prop; it should return 0 like the fall-through path.

`prop2Ptr != NULL` is worth folding in while you're there. A NULL array with a non-zero count is one of the ways you get the garbage index in the first place.

src/port/Patches/PerfPatches.cpp
JeodC: I raised this one because the rest of the file is replacement math, and what you added here is just the warn logging. I have no preference though.

Malkierian: I can't think of a better place at the moment besides just putting it in Engine, but I'll put a comment on it to find a better place later.

PR #312, opened 2026-08-01
Fix rando savestate restore crashing and duplicating junk items

JeodC:
Resolves a crash with junk spawns in rando. `isSaveState` was only held across a single `IsActorWhitelisted` call, but the spawn it governs runs later; `OnActorSpawn` re-evaluated with the flag clear, reached the opposite verdict, and returned NULL. Now it's held for all of `actors_applyFromSavestate`, `OnLoadActorSaveState` resolves the check before deciding what to do with it so the existing "already spawned" cancel covers junk too, and two nullable spawn results are guarded.


JeodC:
@Caladius If this isn't up to your spec then please rework. Does resolve the crash.

PR #327, opened 2026-08-01
Fix CCW jiggy podium visibility

### LINKED ISSUE #326: CCW Podium visible too early
Started a fresh file (slot 1), 100%ed MM and TTC, went to check the CCW painting room out of curiosity, and the jiggy podium was already there. It doesn't do anything if you stand on it, but it's obviously not supposed to be there yet.

Reported by Cyberlink420

Image

JeodC:
Reworks the anchor sync to be like boggy3, where vanilla path runs when anchor isn't connected at all.

Resolves #326


Malkierian:
Just noting clang failed on this one.

Malkierian (approved):

PR #370, opened 2026-08-03
[RANDO] Spoiler Generation Option

Caladius:
Adds Generating and Loading from spoiler logs.


JeodC:
```
src/port/Rando/Spoiler/File.cpp:2:10: fatal error: port/ui/Notification.h: No such file or directory
src/port/Rando/Spoiler/Generate.cpp:4:10: fatal error: port/ui/Notification.h: No such file or directory
```
Change `ui` to `UI` in these paths. Case sensitivity.

Issue #341, opened 2026-08-02
[ROMHACK] New Horizons cannot be completed


CyberLink420:
Tested on 1.0.1. In normal gameplay, passing through the 90 note door is supposed to access the train station, which is required to progress. However, in Lighthouse it teleports the player back to Banjo's house, meaning 25% of the hack is fully inaccessible.

[Save file attached for testing](https://github.com/user-attachments/files/30626316/file1.json)

JeodC:
Malkierian can we leave issues like this open with a romhack label for keeping track?

CyberLink420:
Encountered an identical issue in Banjo-Dreamie where the very first door sends the player back to the spawn point, making it impossible to leave the first room. Both issues are almost certainly tied to the same root cause.

PR #380, opened 2026-08-03
Batch of romhack bugfixes

### LINKED ISSUE #377: [ROMHACK] Nightbear Before Christmas: Switch automatically triggers
Tested on 1.0.1 (Windows, OpenGL). When starting the game, the door to Grunty's Lair immediately opens (which normally requires hitting a switch), allowing players to skip the entire prologue section. [Video demonstration](https://github.com/user-attachments/assets/dfa728d3-abe1-475a-81f2-4d2c5d10c07e)

JeodC:
- Cut-Throat Coast: call VB event and exorcise the ghost from the rock (keep it from rising up on map reload)
- Add a flag tracer to dev tools
- Anchor: home-scope the shared puzzle mirrors; each puzzle id only operates in its actor's vanilla map
- Fix bottles tutorial gated on map instead of level (map originally used because of a collision with the BGS swamp switch)

Resolves #377. Banjo-Kazooie tracks a lot of small per-map state in numbered slots: "map flag 1", "map flag 0", etc. Those numbers mean completely different things depending on which map you're standing in. In Spiral Mountain, map flag 1 means "talked to Bottles". In Treasure Trove Cove, it means "Blubber got his second gold bullion". NBC moves Blubber out of Treasure Trove Cove and into Spiral Mountain. So Blubber woke up, read "map flag 1", saw it was set (because Bottles had just set it) and concluded he'd been given a bullion. He published that, read it back, and set map flag 0 to match. In that map, a repurposed gate reads map flag 0 to mean "the switch was pressed", so up it went, one frame after Bottles finished talking.

- It happened in single player. The helper kept its bookkeeping locally even with nobody connected, so the corruption needed no network at all.
- It affects any romhack.

Each store entry is written by a specific vanilla actor that lives in exactly one known map. Every puzzle id is now scoped to that home, and outside it the whole mirror is inert. Nothing publishes, nothing reads back, and incoming packets and team-state snapshots for foreign maps are dropped (so a peer on an older build can't push them back in). NBC's Blubber runs in a map that isn't Treasure Trove Cove, so his mirror simply doesn't exist there, connected or not.

What that means in practice:

- Vanilla is untouched in every mode, these actors only ever run in their home maps.
- Single player keeps all of its bookkeeping.
- Romhack co-op keeps live sync for every puzzle the hack didn't relocate: your partner does see your third croctus light up.
- Gold bullions and the bear cubs' presents pair respawn-suppression with a credit-back; those are scoped together per kind, so a suppressed item is always credited and a relocated one simply respawns like on hardware.
- The tutorial choice is homed by level rather than map, for the same reason as the abilityprogress fix.

Finding all this is made easy by Flag Tracer, which is now in Dev Tools -> Gameplay Tools -> Monitoring. Tick a few boxes and play, and it's all logged with trace:
```
[Flag] ===== map 0x6A -> 0x1 (level 0xB, exit 1) =====
[Flag] Map           0x1 SM_SPECIFIC_FLAG_1_TALKED_TO_BOTTLES -> ON
set by:
chSmBottles_update (smbottles.c:487)
gsworld_update (gsworld.c:358)
```


JeodC:
The latest commit resolves #341 where we wrongly applied BKCF to `warp_lairEnterMMLobbyFromSMLevel`. That blocked the New Horizons warp to the train station. It is now unblocked.

src/lair/lairspawnqueue.c
```diff
@@ -1086,24 +1091,83 @@ void func_80387730(Actor *this) {
...
+    return NULL;
+}
+
+// [port] Anchor: one-shot cues for a teammate's file-progress flag.
+void port_progressFlag_remoteCue(s32 progressFlag) {
+    Actor *actor;
+
+    port_notedoor_remoteOpen(progressFlag);
+    port_leveldoor_remoteOpen(progressFlag);
+    port_breakable_remoteBreak(progressFlag);
```
Malkierian, line 1117: Are these allowed to call in series simply because there's no overlap in the progress flags between the three types?

Malkierian: And they early return themselves if they're not in their own ranges?

JeodC: Yes, exactly. The three are disjoint slices of `file_progress_e`, and each one filters on its own range and no-ops on anything outside it, so at most one of them (or the switch below) ever acts on a given flag:

`port_notedoor_remoteOpen`: 0x3A-0x45, explicit range guard
`port_leveldoor_remoteOpen`: 0x28-0x30 and 0xE2, switch with default: break
`port_breakable_remoteBreak`: 0xA5, 0xC2-0xC5, 0xC8–0xCE, table lookup that returns when there's no match
the switch here, 0x1E–0x21, 0xA1

I verified all four sets pairwise against the enum and they don't intersect. It wasn't obvious at the call site though, partly because 0x28-0x30 are bare literals. The enum has no names for them, it jumps straight from `FILEPROG_27_LAIR_WATER_LEVEL_3` to `FILEPROG_31_MM_OPEN`.

JeodC:
The Gruntch and Santa's Village are now fully ported, and I added launch arg handling for romhacks. `-hack <romhack>` where `-hack` is aliased to the `~romhacks` directory (by using `GetPathRelativeToAppDirectory("mods")`). This satisfied #386.

Issue #385, opened 2026-08-03
Mumbo Untransform Checkbox does nothing.


Teraunce:
- [x] I have checked the known issues and searched Github to ensure this issue has not already been reported.
- [x] This is not a feature request, I know those belong on Discord/Github Discussions.

### Description of the bug:
The "No Mumbo Untransform" checkbox does nothing, at least in Rando.

### Can you reproduce? If so please list the steps:
- [ ] No, I cannot reproduce this bug
- [x] Yes, here are the steps to reproduce the bug:

1. Be in a Transformed State.
2. Walk over any vanilla untransform location.
3. Ignore the warnings.

### Expected behavior:
Given what the option float text says Banjo should never see warnings or be untransformed.

### Screenshots or videos:
[![Video Title](http://img.youtube.com/vi/EPUQSy9mvFY/0.jpg)](http://www.youtube.com/watch?v=EPUQSy9mvFY)
https://youtu.be/EPUQSy9mvFY

### System Information:
- OS: Win 11
- GFX Backend: DirectX
- Game Version: 1.0.1
- Rom Version: 1.0U

Caladius:
Fixed in https://github.com/HarbourMasters/Lighthouse/pull/388

Teraunce:
Probably fixed thanks to that commit. Closing it myself so it doesn't clutter the issues tab.

PR #388, opened 2026-08-04
Fix No Mumbo Detransform

Caladius:
Moves No Mumbo Detransform to a VB Hook. Tested that it fires off correctly and no longer uses an expensive OnGameFrameUpdate call.


Caladius:
Fixes https://github.com/HarbourMasters/Lighthouse/issues/385

PR #395, opened 2026-08-04
Fix input viewer

aMannus:
Diff is huge but it simply removed the cvar check around the whole draw function. It still toggles just fine with the button because it's the registration of the window that already handled that.

Also found a 2ship icon still in the assets folder so figured I'd remove that while I was in there, lol.


src/port/UI/LighthouseGui.cpp
```diff
@@ -201,7 +201,7 @@ void SetupGuiElements() {
     mInputViewer = std::make_shared<InputViewer>("gWindows.InputViewer", "Input Viewer");
     gui->AddGuiWindow(mInputViewer);
 
-    mInputViewerSettings = std::make_shared<InputViewerSettingsWindow>("gWindows.InputViewerSettings",
+    mInputViewerSettings = std::make_shared<InputViewerSettingsWindow>("gOpenWindows.InputViewerSettings",
```
Malkierian: There's a macro for that prefix.

Malkierian:
> Diff is huge but merely moved CVar

Well, thank goodness for this then
image
XD

PR #396, opened 2026-08-04
Disable Note Retention during parades

### LINKED ISSUE #342: Extra lives granted during cast parades
Tested on 1.0.1. While viewing either of the cast parades, whenever the camera cuts to a new world, it will grant Banjo an extra life if all the notes in that world have been obtained. [Here's a video sample.](https://github.com/user-attachments/assets/3b09fafe-fd3f-47a3-add3-5b0b7537bf79)

Based on my observations, it seems to have to do with how notes are recorded when note saving is turned on. When hitting witch switches, I sometimes notice the note counter appearing in the corner and quickly counting up from zero. to whatever my previous note count was. Something similar may be happening during the cast parades, triggering the 100-note bonus life as it instantly counts up to the max for each completed world.

JordanLongstaff:
Fixes #342

### DIFF
--- a/src/port/Enhancements/Retention/NoteRetention.cpp
+++ b/src/port/Enhancements/Retention/NoteRetention.cpp
@@ -139,6 +139,11 @@ int32_t countCollectedForLevel(int32_t levelId) {
     return total;
 }
 
+bool isInParade() {
+    return volatileFlag_get(VOLATILE_FLAG_1F_IN_CHARACTER_PARADE) ||
+           volatileFlag_get(VOLATILE_FLAG_C1_IN_FINAL_CHARACTER_PARADE);
+}
+
 } // namespace
 
 extern "C" void port_noteRetention_getSizeAndPtr(int32_t* size, uint8_t** addr) {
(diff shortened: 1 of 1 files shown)


JeodC:
Note Retention wasn't really my domain. I saw you pinged me on that issue, and I did made two recent changes with https://github.com/HarbourMasters/Lighthouse/pull/283 and https://github.com/HarbourMasters/Lighthouse/pull/336. I believe they are related, but not altogether the solution as they hide the hud but you would still be granted an extra life. Then there's [#293](https://github.com/HarbourMasters/Lighthouse/pull/293/changes#diff-a2fa5d9186cf3297f52c6d72094d7a2b928deedd21577c2639570f99cc718389) which you linked in the issue, which was my attempt at ensuring retention stayed out of attract demos and bottles bonus (during trying to fix syncs, the thought was they might be affecting the demo's deterministic seed). If that was done incorrectly, then this is a good place to fix it.

I'll cc @Malkierian as they wrote the retention.

Malkierian:
I made some significant changes to how note retention works in a PR I'll be making soon. I don't think this will be an issue, but I'd like to have the parade tested with that PR before merging this in case I already incidentally fixed it with the other issues I fixed.

Malkierian:
@JordanLongstaff The other Anchor port is merged. Go ahead and test latest develop-splitrock to see if this is necessary. If it is, this will be the last PR before Charlie.

JeodC:
> @JordanLongstaff The other Anchor port is merged. Go ahead and test latest develop-splitrock to see if this is necessary. If it is, this will be the last PR before Charlie.

It is.

JordanLongstaff:
Thanks for doing that so I don't have to @JeodC 👍

PR #399, opened 2026-08-04
feat(android): add SDL/GLES launcher

pacoa-kdbg:
## Summary
- Add an Android SDLActivity/Gradle launcher and Android NDK build path.
- Package extraction assets and the prebuilt O2R resource archive.
- Include Android portability fixes for Bionic, storage bootstrap, and host-only tooling.

## Dependency
This port depends on the accompanying libultraship Fast3D ARM64 tagged-pointer fix: Kenix3/libultraship#1228.

## Validation
- AI-generated implementation.
- Built, installed, and device-tested on an AYANEO Pocket S Mini (Adreno 740, Android 13).
- User subsequently performed manual testing on the same device.

## Notes
The current submodule pin is from the development fork while the libultraship fix is reviewed upstream; it should be updated to the upstream commit after that PR merges.


JeodC:
Can Gradle be made to make a copy of the root `assets` dir at build time? When we pull upstream decomp changes that carry enum modifications, those assets yaml entries will be edited and updated. Maintaining them in several different places is something we want to avoid.

pacoa-kdbg:
Implemented in 77cfbd52.

- Gradle now `Sync`s the root `assets/` directory into generated APK assets at build time.
- Removed the checked-in Android asset-tree duplicate.
- Verified a clean Android Gradle build succeeds and the generated `assets/yaml/us/rev0/assets.yaml` SHA-256 matches the root source asset.

JeodC:
Thanks, you should be able to do the same with `config.yml` and maybe `lighthouse.o2r` since that may also change down the road if LUS refactors its shaders again or we modify anything custom to pack inside (custom actors, objects, models, sprites etc).

pacoa-kdbg:
Follow-up implemented in d9f2235d. `config.yml` now also syncs from the repository root into generated APK assets; its checked-in Android copy was removed.

Verified `:app:assembleDebug` succeeds and generated `config.yml` is byte-identical to the root source.

`lighthouse.o2r` is currently a generated artifact rather than a root source file; I am keeping it separate for now rather than introducing a second artifact copy. A build-time Torch generation task can be added once the desired host-tool bootstrap approach is agreed.

pacoa-kdbg:
Regarding `lighthouse.o2r`: it remains a checked-in Android APK asset for now because it is a generated archive, not a canonical root source file. Its inputs are root `port/` plus `libultraship/src/fast/shaders/`.

The proper follow-up is a Gradle task that bootstraps/runs host-native Torch, stages those inputs, and writes the O2R into generated APK assets; then the checked-in binary can be removed. I kept that separate because host-Torch bootstrapping needs to be designed to work predictably for upstream CI and developer builds, independently of the Android NDK build.

Malkierian:
Sorry it took so long to get here, but our policy for now is that we don't want to merge other platform support into main, as we have a hard enough time maintaining the three PC builds. Go ahead and make a thread in lh-chat in the server if you'd like to release it.

PR #404, opened 2026-08-05
Fix WorldStates for the 1000th time.

Caladius:
Fixes the BGS Jiggy Switches and Boggy States by adding a JiggyCollected guard.


Teraunce:
Causes Double the Boggys for Double the Fun?
image
Boggy 1 started the race immediately before the countdown started while Boggy 2 did the countdown. Thankfully Boggy 2 was the one I had to beat.

JeodC:
That would be Boggy2 and Boggy3 actually.

Teraunce:
both races' sets of slalom gates spawned during the first race.

aMannus:
Superceded by https://github.com/HarbourMasters/Lighthouse/pull/423

PR #423, opened 2026-08-06
Rewrite parts of rando, code cleanup

JeodC:
`randoCollectionId` holds the **item** placed at a location, but the score listeners matched on it to answer **world-state** questions. `jiggyscore_isCollected(N)` returned the state of whatever check received jiggy N. Which checks broke was seed luck; on a test save, 14 of 31 world-state jiggy ids and 9 of 24 honeycombs were already wrong.

- `OnIsJiggyScoreCollected` / `OnIsHoneycombScoreCollected` now resolve via `GetCheckByJiggyId` / `GetCheckByHoneycombId`. Supersedes the hand-written id special-cases (GV water pyramid stays, draining needs the jiggy *and* the token).
- Split inventory from world state: the per-level tallies, `chjiggy`/`chHoneycomb` despawn tests and Anchor dedup read new raw accessors; jiggy/honeycomb counts and item sync still work.
- BGS jiggy switches: `actorArray_findActorFromActorId` grabbed an unrelated check, played the destroy camera on it and left it `collisionOff` forever. Now uses `FindActorByRandoCheckId` and bails if already obtained.
- Removed `Boggy.cpp`, the original code path drives Boggy.
- Removed `Gobi.cpp`, the original code path drives Gobi.
- FP xmas tree switch works again.
- New files: no 100-note jingle / Bottles dialog / free extra life, and Mumbo isn't pre-paid for the bee.
- Skip Spiral Mountain Tutorial disabled under rando (`DISABLE_FOR_RANDO`); its rando-side grant of 12 SM checks removed. Rationale: under Rando, Spiral Mountain is part of the challenge. Also prevents potential future bugs. Stay away from Tutorial Bottles and Tutorial Bottles can't hurt you.
- Fixed an OOB write past `RANDO_SAVE_FLAGS` and inverted `!IS_RANDO && !OPTION` guards.
- New: rando saves show Cheato on the file select.

**Tested** on a 900+/1200 seed: 15 gates where old and new disagree (5 of the 12 reported), BGS switch, FP Boggy chain, xmas tree.

**Merging:** #404 is superseded. Conflicts with `aMannus/custom-collectible` in `WorldState.cpp` (take mine + `obtained`->`eligible`) and `Boggy.cpp`/`Gobi.cpp` (take the delete).


include/functions.h
```diff
@@ -309,9 +309,11 @@ s16 *picturebox_getColorBuffer(void);
 // --- core2/jiggyscore.c ---
 u32 jiggyscore_isCollected(enum jiggy_e jiggy_id);
 int jiggyscore_isSpawned(enum jiggy_e jiggy_id);
+u32 port_jiggyscore_isCollectedRaw(enum jiggy_e jiggy_id);
 
 // --- core2/honeycombscore.c ---
 bool honeycombscore_get(enum honeycomb_e indx);
+bool port_honeycombscore_getRaw(enum honeycomb_e indx);
```
aMannus, line 315: We should probably not let this block the PR in the interest of getting this fixed asap, but these will have to be replaced with proper hooks gated behind IS_RANDO down the line.

src/port/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -79,10 +80,17 @@ bool failSafeTrigger = false;
...
 std::vector<RandoCheckId> jinjoCheckIds;
 
+static void SetItemCountSilently(item_e item, int32_t value) {
+    D_80385F30[item] = value < 0 ? 0 : value;
+    if (item == ITEM_1C_MUMBO_TOKEN) {
+        D_80385F30[ITEM_25_MUMBO_TOKEN_TOTAL] = D_80385F30[item];
+    }
+}
+
 void UpdateSaveDataItemCounts(PlacedItemCounts itemCounts) {
```
aMannus: What's the reason this function was made instead of using the vanilla one that already exists?

src/port/Rando/MiscBehavior/FileSelect.cpp
```diff
@@ -0,0 +1,38 @@
...
+    func_803152C4(zoombox);
+    zoombox->unk1A4_14 = 0;
+    zoombox->unk1A4_13 = 0;
+    zoombox->unk17C = 0.0f;
+}
+
+void Rando::MiscBehavior::InitFileSelectBehavior() {
+    REGISTER_LISTENER(OnFileSelectPortrait, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        OnFileSelectPortrait* ev = (OnFileSelectPortrait*)event;
+
```
aMannus, line 23: I know this is following rando's structure that already exists, but this needs to be converted by registering with ShipInit down the line. Just like the port_ functions, this probably shouldn't block the PR, but noting it here so we have reference later.

aMannus (commented):
This looks great overall! Couple of notes and one question from me

src/port/Rando/Logic/GlitchlessLogic.cpp
JeodC: `item_adjustByDiffWithoutHud` only suppresses the HUD print; the switch(item) block below it still runs. The solver walks these counts as bookkeeping while it decides placements, on the file select before the seed has started, so taking `ITEM_C_NOTE` through 100 fired the 100th-note fanfare, awarded a life and popped Bottles' high-score dialog.

aMannus (approved):

src/port/Rando/Logic/GlitchlessLogic.cpp
Caladius: Why do we need a function for this when this function is basically just setting a value in an array? Can't we just set the value directly instead in a one liner instead of fluffing 6 lines of code for it?

src/port/Rando/ObjectBehavior/Switch.cpp
```diff
@@ -39,9 +41,12 @@ void Rando::ObjectBehavior::ModifySwitchBehavior(int32_t switchActorId) {
         return;
     }
 
-    RandoSaveCheck randoSaveCheck = Rando::Logic::GetShuffledObject(randoCheckId);
-    actor_e actorId = (actor_e)Rando::StaticData::Items[randoSaveCheck.randoItemId].actorId;
-    Actor* findActor = actorArray_findActorFromActorId(actorId);
+    // Don't play missed switch scene if the check's been obtained
+    if (Rando::Logic::GetShuffledObject(randoCheckId).obtained) {
```
Caladius: Doesn't need to be in this PR but a note for later to get rid of these that are just checking 1 thing and instead use RANDO_SAVE_CHECKS[randoCheckId] directly.

src/port/Rando/MiscBehavior/OnFileLoad.cpp
```diff
@@ -39,7 +39,6 @@ void Rando::MiscBehavior::OnFileLoad() {
                 Rando::Logic::GenerateShufflePool(saveData);
                 Rando::Logic::GrantStartingLoadout();
                 Rando::Logic::GrantFileProgressFlags();
-                Rando::Logic::GrantSpiralMountainChecks();
```
Caladius, line 42: Where are you applying the Spiral Mountain checks with this function removed? I don't see its new workflow to give those checks if SkipSpiralMountain is enabled.

JeodC: SkipSpiralMountain is disabled for rando modes.

> - Skip Spiral Mountain Tutorial disabled under rando (DISABLE_FOR_RANDO); its rando-side grant of 12 SM checks removed. Rationale: under Rando, Spiral Mountain is part of the challenge. Also prevents potential future bugs. Stay away from Tutorial Bottles and Tutorial Bottles can't hurt you.

src/core2/ch/jiggy.c
```diff
@@ -146,7 +146,8 @@ void chjiggy_update(Actor *this) {
                 local->id = getJiggyId(this);
             }
             
-            if (jiggyscore_isCollected(local->id)) {
+//          if (jiggyscore_isCollected(local->id)) {
+            if (port_jiggyscore_isCollectedRaw(local->id)) {
```
Caladius: I don't know how I feel about Vanilla code being replaced with a function whose code lives inside a Rando file.

If we need to replace this I think the function should live somewhere in our global functions file or maybe ShipUtils?

src/core2/ch/honeycomb.c
```diff
@@ -105,7 +105,8 @@ void chHoneycomb_update(Actor *this){
         if( this->marker->id == MARKER_53_EMPTY_HONEYCOMB){
             if(local->uid == 0)
                 local->uid = (!this->unk44_2)? D_8037DDC0 : func_802C9C40(this);
-            if( honeycombscore_get(local->uid)
+//          if( honeycombscore_get(local->uid)
+            if( port_honeycombscore_getRaw(local->uid)
```
Caladius: same as my comment on port_jiggyscore_isCollectedRaw

src/port/Rando/MiscBehavior/OnFileLoad.cpp
Caladius: Personally don't think it should be disabled since we already know it can work, just needed an iteration. We also shouldn't assume on behalf of the player?

JeodC: Re-enable it with v2? When I tested this branch with it still turned on it would grant me all of the vanilla tutorial abilities instead of the checks, and I just didn't feel like chasing it down right now given how much tutorial bottles has bit.

src/port/Rando/ObjectBehavior/Switch.cpp
JeodC: Nah I'll do it now. Good shout.

src/port/Rando/MiscBehavior/OnFileLoad.cpp
Caladius: Non blocking for the PR but this being disabled for Rando will be a blocker for Rando V2 as a whole.

JeodC: When the milestone for v2 is live I'll create an issue to attach to it as a follow-up, then. The whole idea of this was to create less problems for v2 after all.

src/port/Rando/Logic/GlitchlessLogic.cpp
JeodC: Because it has a negative clamp guard in it. If the solver doesn't need that I'll reduce it to direct sets.

src/core2/ch/jiggy.c
JeodC: Moved to `ShipUtils.cpp` as suggested, and both of these call sites no longer use them at all.
- `ch/jiggy.c` and `ch/honeycomb.c` now go through `VB_COLLECTABLE_ALREADY_HELD`. The default is the vanilla expression, so a vanilla-spawned collectable answers exactly as before; rando overrides only when `marker->randoCheckId` is set and answers from `RANDO_SAVE_CHECKS`, so it never touches the bitfield.
- `jiggyscore_leveltotal` is back to pure decomp. `VB_JIGGYSCORE_LEVEL_TOTAL` already existed for romhacks, rando just registers it now. `honeycombscore_get_level_total` got the matching `VB_HONEYCOMBSCORE_LEVEL_TOTAL`.
- Declarations moved out of `functions.h` into `ShipUtils.h`, so decomp doesn't see them either.

Decomp is now down to zero `port_*` score calls. The two accessors are port-only, with two consumers: Anchor's three dedup guards, and rando's two tally handlers. I would like to put them in Anchor, but then Rando would be including an Anchor header for them.

I did look at making them Anchor-only by having the tally handlers count from `RANDO_SAVE_CHECKS` instead, but that trades the authoritative record for a parallel one: the bitfield is what `jiggyscore_setCollected` writes on every pickup, whereas in Anchor a remote collect arrives as two packets (`COLLECT_ITEM` writes the bitfield, `SET_CHECK_STATUS` writes the check), so any drift would show up as a wrong jiggy count.

Caladius (approved):

PR #460, opened 2026-08-09
Fix event priority mismatch in LoadGlobalData.

JeodC:
Also fix SNS to use OnGameStart; the event priority mismatch had broken persistent bottles bonus.

### DIFF
--- a/src/port/Save/SaveManager.cpp
+++ b/src/port/Save/SaveManager.cpp
@@ -926,7 +926,7 @@ void SaveManager_Init() {
     // Decomp clears global arrays (e.g. gCompletedBottlesBonusGames) just before
     // gameFile_load fires OnGameLoad. Restore them from global.json here before
     // other OnGameLoad listeners read them.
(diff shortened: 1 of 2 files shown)


src/port/Save/SaveManager.cpp
```diff
@@ -926,7 +926,7 @@ void SaveManager_Init() {
     // Decomp clears global arrays (e.g. gCompletedBottlesBonusGames) just before
     // gameFile_load fires OnGameLoad. Restore them from global.json here before
     // other OnGameLoad listeners read them.
-    REGISTER_LISTENER(OnGameLoad, EVENT_PRIORITY_HIGH, [](IEvent* event) { LoadGlobalData(); });
+    REGISTER_LISTENER(OnGameLoad, EVENT_PRIORITY_LOW, [](IEvent* event) { LoadGlobalData(); });
```
Malkierian, line 929: Is this meant to be low, or is this a side effect of the inverted priority order you were talking about? If the latter, a note should be made to change it back when we fix that in LUS.

JeodC: Side effect. I would wait until we do LUS 2.0 migration to fix it though.

Issue #489, opened 2026-08-11
[Rando] Multiple Blue Egg Items


mattman107:
https://github.com/HarbourMasters/Lighthouse/blob/052d7bd24178078cfb4d6648c580c32693eadce5/src/port/Rando/StaticData/Items.cpp#L23

https://github.com/HarbourMasters/Lighthouse/blob/052d7bd24178078cfb4d6648c580c32693eadce5/src/port/Rando/StaticData/Items.cpp#L116

There are two items in the Items map that map to the string "Blue Egg". AP is likely to use this to map to what item to give to the player, so it will need to be unique.

Would there be any issue if the Stop N Swap items became something like this instead (feel free to suggest something better too):

Blue Egg -> Stop N Swap Blue Egg
Cyan Egg -> Stop N Swap Cyan Egg
Green Egg -> Stop N Swap Green Egg
Pink Egg -> Stop N Swap Pink Egg
Red Egg -> Stop N Swap Red Egg
Yellow Egg -> Stop N Swap Yellow Egg
Ice Key -> Stop N Swap Ice Key

Caladius:
Putting my suggestion in to use the Notification message for the name instead:

Mystery Blue Egg
Mystery Cyan Egg
etc.

Stop N Swap Blue Egg feels too long

Teraunce:
SnS Blue Egg?

mattman107:
#491 was merged so I'm closing this.

PR #491, opened 2026-08-12
Rename sns item names

mattman107:
For #489


Teraunce:
I'm sad you didn't go with my suggestion of "SnS 'insert item name here'" as it tells you exactly what collectible set it's for, but starting with "Mystery" works as well.

Caladius (approved):

PR #494, opened 2026-08-12
Contribus

PurpleHato:
image

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9160430444.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9159125849.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9158796607.zip)

### DIFF
--- a/src/port/UI/LighthouseMenuSettings.cpp
+++ b/src/port/UI/LighthouseMenuSettings.cpp
@@ -78,7 +78,7 @@ static const std::unordered_map<int32_t, const char*> saveConvertSlotLabels = {
 
 static int32_t sAppliedControlScheme = -1;
 
-// Contributors sorted by commit count (GitHub handle preferred; real name where no handle is known)
+// Contributors (GitHub handle preferred; real name where no handle is known)
 static std::vector<std::string> contributors = {
     "Banjo Decomp", "Caladius",        "JeodC",        "Malkierian",   "ProxySaw",    "PurpleHato",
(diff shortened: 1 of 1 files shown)


Caladius:
The top section with the Port name, Branch, and Commit were moved to the ( ? ) button at the top right. I also don't know if this is something we want to add maintenance work for.

PurpleHato:
> The top section with the Port name, Branch, and Commit were moved to the ( ? ) button at the top right. I also don't know if this is something we want to add maintenance work for.

I never noticed that button until now, thanks Cal;
Well it's not something that **needs** updating everytime someone is gonna contribute + we have easy access on new contributors on releases on both github and website so it's not tooo much maintenance personally, it's something I can do from time to time for sure

Caladius:
>Well it's not something that **needs** updating everytime someone is gonna contribute + we have easy access on new contributors on releases on both github and website so it's not tooo much maintenance personally, it's something I can do from time to time for sure

Thats the point I'm trying to make, someone has to keep up with this added effort.

PurpleHato:
> 
> Thats the point I'm trying to make, someone has to keep up with this added effort.

Can do that when it's necessary, if you think it's too much, I can also just close it, just thought it would be a cool idea to have it asap before doing the opposite and needing to chase duplicate accounts etc due to mail difference and all but keeping the same alias via the API (easier for sorting that out)

JeodC:
If there were a way to automate it at build, sure, but I would not like to sign up for additional maintenance.

PR #505, opened 2026-08-14
Implement Anchor GIVE_ITEM receive handler

chidmdxx:
`HandlePacket_GiveItem` is a commented-out stub carried over from Ship of Harkinian, so a `GIVE_ITEM` packet that arrives from a remote item source is parsed and then dropped. This implements the receive side against the randomizer item queue so the item actually lands in the player's save.

This replaces #504, which changed the packet contract. That turned out to be unnecessary, so this version keeps the existing contract and is much smaller.

## What changed

One file, `src/port/Network/Anchor/Packets/GiveItem.cpp`. `Anchor.h` is untouched.

`SendPacket_GiveItem` keeps its `(u16 modId, s16 getItemId)` signature and the payload keeps its `modId` and `getItemId` fields, so any existing sender keeps working. The send path is otherwise unchanged apart from dropping commented-out OoT logic (ice traps, the Ganon boss scene) and the unused `incomingIceTrapsFromAnchor` global that only that dead code referenced.

## Packet contract

`modId` selects which item table `getItemId` belongs to. It keeps the meaning it had in the packet this file inherited:

| `modId` | meaning |
|---|---|
| `0` | vanilla item table, not implemented here, dropped |
| `1` | randomizer item table, `getItemId` is a `RandoItemId` |

Only the randomizer table is implemented. A packet with any other `modId` is dropped rather than guessed at, which leaves room for the vanilla table later without changing the wire format.

## Validation

The handler uses the same guards as the other world packets: `IS_RANDO`, `IsSaveLoaded()`, and `roomState.syncItemsAndFlags`.

`getItemId` is read into an `s32` and checked with `getItemId <= RI_UNKNOWN || getItemId >= RI_MAX` before anything uses it. Reading into a wider type first means an out of range value is rejected instead of wrapping into a valid id. The check matters because `ItemQueue` and `StaticData` index `Items[]` with this value directly, so a malformed or stale packet must not reach them.

Missing fields fall back to defaults that fail validation, and a field of the wrong JSON type throws inside the `try` block in `Anchor::ProcessIncomingPackets`, so neither case reaches the item queue.

There is no explicit `IsGlobalRoom()` guard because the existing code already covers it. `HandlePacket_UpdateRoomState` forces `syncItemsAndFlags` to `0` when `IsGlobalRoom()`, and `GIVE_ITEM` is not in `AllowedWithoutGameSync`, so `OnIncomingJson` drops the packet before dispatch. Adding a second check would only restate that.

## Idempotency

A grant is a delta, not a set. Counts increment and the next free jiggy or honeycomb slot is taken, so nothing here deduplicates. The same packet delivered twice grants the item twice, and a grant cannot be taken back. At-most-once delivery is the sender's responsibility.

## Notifications

`ItemQueue::SendNotification` self gates on the randomizer notification CVar, so a remote grant is announced on the same terms as a local pickup.

## Testing

`clang-format` 14 is clean on the changed file. The repo has no test harness, so nothing was added there.

I have not compiled this locally. The machine I am on has no C++ toolchain installed, so I am relying on the CI build. The change is one translation unit and only calls existing public APIs (`ItemQueue::GiveItem`, `ItemQueue::SendNotification`), so please flag anything the build catches and I will fix it.

### DIFF
--- a/src/port/Network/Anchor/Packets/GiveItem.cpp
+++ b/src/port/Network/Anchor/Packets/GiveItem.cpp
@@ -1,33 +1,46 @@
 #include "port/Network/Anchor/Anchor.h"
 #include <nlohmann/json.hpp>
 #include <libultraship/libultraship.h>
-#include "port/UI/Notification.h"
 
-#include "functions.h"
-// extern PlayState* gPlayState;
+#include "port/UI/cvar_prefixes.h"
+#include "port/Rando/Rando.h"
+#include "port/Rando/ItemQueue/ItemQueue.h"
 
 /**
  * GIVE_ITEM
  *
- * unimplemented
+ * Hands a single item to this client. SET_CHECK_STATUS mirrors a check a teammate already
+ * obtained; GIVE_ITEM instead says "this item is yours now", which is what an external item source
+ * needs when it decides an item belongs to this player.
+ *
+ * Payload (unchanged from the packet this file inherited, so existing senders keep working):
+ *   modId      u16  which item table getItemId belongs to
+ *                     0  vanilla item table, not implemented here, ignored
+ *                     1  randomizer item table, getItemId is a RandoItemId
+ *   getItemId  s16  the item id, read according to modId
+ *
+ * Only the randomizer table is implemented. A packet with any other modId is dropped rather than
+ * guessed at.
+ *
+ * Safety:
+ *   - Guarded like every other world packet: rando save loaded and room item sync on. That also
+ *     covers the public global room, where HandlePacket_UpdateRoomState forces syncItemsAndFlags
+ *     to 0 and OnIncomingJson then drops GIVE_ITEM before it reaches this handler.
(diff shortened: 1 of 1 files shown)


Malkierian:
Anchor has no need of this. If Archipelago gets implemented, it will have its own system, but SetCheckStatus handles evaluating co-op check resolution locally, thus granting the items from the checks and despawning locally. GiveItem was never even sent, let alone received and ignored. It was intentionally non-operational, and I had no plans of implementing Anchor-centric multiworld functionality.

Also, the burden is not on us, but on you, to test changes you attempt to have an AI agent make, and you have to be able to understand/explain what it's doing. Since you are incapable of doing so due to not having a ROM, this would not be evaluated further anyway.

chidmdxx:
completly fair, honestly thought the testing framework that build around it should've been picked up before sending an untested PR...

PR #522, opened 2026-08-17
Fix Chimpy softlock in Anchor

JeodC:
This attempts to address the Chimpy anchor softlock.

- Holds `jiggyscore_isSpawned(JIGGY_9)` false for the couple of seconds a local delivery is playing out; the orange throw is the only way in, Chimpy walking off or the map going away the only way out
- Rando's answer for `RC_MM_JIGGY_CHIMPY` drops the `.obtained` term that `AdoptRemoteCheck` flips mid-map
- Refuses a second `Jiggy #9` when the team already has one, so finishing the scene late can't double up

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9304518551.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9304248030.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9303954726.zip)

### DIFF
--- a/src/port/Enhancements/Fixes/GameFixes.cpp
+++ b/src/port/Enhancements/Fixes/GameFixes.cpp
@@ -164,6 +164,43 @@ void RegisterChimpyStumpRumble_Init() {
     });
 }
 
+// Chimpy softlock prevention
+static bool sChimpyDeliveryInFlight = false;
+
+// Lazily expires the in-flight flag.
+static bool ChimpyDeliveryStillInFlight() {
+    if (sChimpyDeliveryInFlight) {
+        sChimpyDeliveryInFlight =
+            gsworld_getMap() == MAP_2_MM_MUMBOS_MOUNTAIN && !mapSpecificFlags_get(MM_SPECIFIC_FLAG_3_CHIMPY_HAS_LEFT);
+    }
+    return sChimpyDeliveryInFlight;
+}
+
+void RegisterChimpySoftlock_Init() {
+    REGISTER_LISTENER(OnCarryThrow, EVENT_PRIORITY_NORMAL, [](IEvent* event) {
+        auto* ev = reinterpret_cast<OnCarryThrow*>(event);
+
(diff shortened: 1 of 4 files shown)


src/port/Enhancements/Fixes/GameFixes.cpp
```diff
@@ -164,6 +164,42 @@ void RegisterChimpyStumpRumble_Init() {
...
+        sChimpyDeliveryInFlight = false;
+    }
+    return sChimpyDeliveryInFlight;
+}
+
+void RegisterChimpySoftlock_Init() {
+    REGISTER_LISTENER(OnIsJiggyScoreSpawned, EVENT_PRIORITY_HIGH, [](IEvent* event) {
+        auto* ev = reinterpret_cast<OnIsJiggyScoreSpawned*>(event);
+
+        if (ev->jiggyId != JIGGY_9_MM_CHIMPY || !ChimpyDeliveryInFlight()) {
```
Malkierian: Should this rather be `== && ==` instead? It seems like this could cause confusion as it is. I think the positive framing is the same functionality, and would be clearer for intent.

JeodC: I reworked the commit, I realized the rando change (second point in description) was a behavior change and not needed for this anyway.

Malkierian: The rework didn't change the reference frame of my question, though.

JeodC: It didn't? I did take the positive framing route.

Malkierian: The changes list still shows `!= || !=`

JeodC: Oh I see it.

PR #530, opened 2026-08-18
Resolve anchor bugs in #525, #526, and #528

### LINKED ISSUE #525: Can duplicate mumbo tokens (and maybe other collectibles) in sub-areas
Seen in a private coop vanilla room. P2 entered a sub-area in MM, and P1 collected a mumbo token in the primary map. When P2 returned, they were able to see and collect the same token for a duplicate.

### LINKED ISSUE #526: Anchor needs documentation on clearing saves in private rooms
In a private anchor room, we began with a vanilla session and then swapped to a romhack in the same roomid. The romhack synced with the vanilla save despite us both using the romhack, and broke the hack entirely. If there is a smooth method for clearing a roomid's synced save data before swapping games, it should be clearly documented.

### LINKED ISSUE #528: Double jiggy HUD in anchor coop during jiggy puzzle completion
When P1 and P2 both stand near a puzzle, the jiggy HUD appears once per player. When P2 completes the jiggy puzzle, a second, slightly offset jiggy HUD and count shows up for P1 until P2's cutscene completes. This is only noticeable if P1 continues to stand near the puzzle to make the first jiggy HUD stay up.

JeodC:
#525: Collectibles duplicated when teammate is in sub-area
- Adds `Anchor_DespawnCollectedCollectibles()`, run from the Anchor map-load hook which fires after `mapSavestate_apply` to despawn any jiggy, honeycomb or mumbo token whose score bit is already set.

#528 - Double jiggy HUD popup
- Anchor's notification is now conditioned on `func_802FADD4(ITEM_2B_UNKNOWN)`, so jiggy total popups are no longer redundant. If a player already has a variant showing, a second one doesn't appear.

#526 - Clearing room state in private rooms
- The server keeps the last team state a team sent and replays it to whoever loads a file when no teammate is online with a save loaded, so a room reused for a different game pasted the old game's save over the new one.
- Team state now carries a `gameSig` (romhack label plus rando seed). A mismatched state is refused instead of applied, along with the packet queue riding on it.
- When nobody on the team is playing, the stale copy is cleared automatically and reseeded from the local file, with a notification. No hunting for an admin-only button.
- When a teammate is playing, their state is left alone and the local player is told instead; outbound team state is also suppressed, so a mismatched client can't poison the room's stored save.
- Room identity had the same staleness: the handshake only seeds it at room creation, so a reused room kept its old romhack/seed. The owner now refreshes it (with a notification) rather than being warned about their own room; non-owners still get the mismatch warning.
- `PrepRoomState` no longer publishes `isRando/seed` while no file is selected. Doing so reset the room's randomizer identity for everyone still in it. Also fixes the pre-existing version of this via the room-settings widgets.
- Owner checks require a server-assigned client id, since both `ownClientId` and `roomState.ownerClientId` default to 0 and would otherwise match before `ALL_CLIENT_STATE` arrives.

There is a greater accessibility problem that will be a separate ticket.

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9739121935.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9738973805.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9738807626.zip)

### DIFF
--- a/src/port/Network/Anchor/Packets/UpdateRoomState.cpp
+++ b/src/port/Network/Anchor/Packets/UpdateRoomState.cpp
@@ -7,6 +7,7 @@
 #include "port/Rando/Rando.h"
 #include "port/UI/LighthouseGui.hpp"
 #include "port/UI/LighthouseModals.h"
+#include "port/UI/Notification.h"
 
 #include "variables.h"
 
@@ -38,8 +39,13 @@ nlohmann::json Anchor::PrepRoomState() {
     payload["shareConsumables"] = CVarGetInteger(CVAR_REMOTE_ANCHOR("RoomSettings.ShareConsumables"), 0);
     payload["isRomHack"] = port_isRomhack();
     payload["romhackName"] = Lighthouse::CurrentRomhackLabel();
-    payload["isRando"] = (bool)IS_RANDO;
-    payload["seed"] = (int32_t)(IS_RANDO ? RANDO_SEED : 0);
+    if (selectedFileNum != DEFAULT_FILE_NUM) {
+        payload["isRando"] = (bool)IS_RANDO;
+        payload["seed"] = (int32_t)(IS_RANDO ? RANDO_SEED : 0);
+    } else {
+        payload["isRando"] = roomState.isRando;
+        payload["seed"] = roomState.seed;
+    }
 
     return payload;
 }
@@ -71,9 +77,19 @@ void Anchor::HandlePacket_UpdateRoomState(nlohmann::json& payload) {
 
     roomState.isRomhack = payload["state"]["isRomHack"].get<bool>();
     roomState.romhackName = payload["state"]["romhackName"].get<std::string>();
+    roomState.ownerClientId = payload["state"]["ownerClientId"].get<uint32_t>();
     const std::string localLabel = Lighthouse::CurrentRomhackLabel();
     if (roomState.romhackName != localLabel) {
-        if (roomState.romhackName != lastWarnedRomhackLabel) {
+        // Only act on ownership the server has actually handed out.
+        if (ownClientId != 0 && roomState.ownerClientId == ownClientId) {
+            roomState.isRomhack = port_isRomhack();
+            roomState.romhackName = localLabel;
+            lastWarnedRomhackLabel.clear();
+            SendPacket_UpdateRoomState();
+            Notification::Emit({
+                .message = "Room now set to " + localLabel,
+            });
+        } else if (roomState.romhackName != lastWarnedRomhackLabel) {
             lastWarnedRomhackLabel = roomState.romhackName;
             std::string msg = "There's a romhack mismatch between your client and the server:\n\n";
             msg += Lighthouse::DescribeRomhackMismatch(port_isRomhack(), localLabel, roomState.isRomhack,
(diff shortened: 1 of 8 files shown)


JeodC:
Reworked this a bit based on discord discussion.

The safety property is now separate from the cleanup. Refusing to apply a team state recorded under a different game is unconditional: every client, regardless of ownership or who's online, and now also when no file is selected yet (fails closed rather than applying something it can't vet). That fixes "play a game, swap to a variant in the same room, variant breaks". Nothing is ever written to your save, so there's nothing to poison even if you save immediately.

Clearing the server's stored copy is a separate decision, and a much more conservative one. A player who wasn't last out is entitled to come back and resume that state. So it's now only ever cleared by the room owner, and never while a teammate is online and playing from it. Non-owners are simply told their file doesn't match. Default is an OK/Cancel modal describing both signatures; a new admin toggle, "Clear Mismatched Team State Without Asking", makes it silent for people who hotswap often.

Two things I'd like a second opinion on:

@Malkierian you mentioned state syncs during slot selection. I traced `selectedFileNum` as being set only in the `OnGameLoad` listener, and `UPDATE_TEAM_STATE` as only arriving in reply to a request we make, so I couldn't find a browse-time path. The fail-closed guard makes it harmless, but if there is one it'd mean the signature can be compared against a half-initialized file.

Room retention: anchor/main has `INACTIVITY_TIMEOUT = 5 * time.Minute` with disconnected clients not being heartbeated, so the clock starts when the last person drops. If the hm64 deployment differs, the "stale state window" reasoning in this PR changes.

Malkierian:
The issue is that OnGameLoad is triggering a forced REQUEST_TEAM_STATE packet, and this gets triggered when swapping to a slot that isn't empty. I think we have a better event to use for that now, though.

Malkierian:
Yeah, OnGameStart.

Malkierian:
As for clear time, yes, it's 5 minutes on the HM server.

JeodC:
Found it, thanks!

src/port/Network/Anchor/Packets/UpdateRoomState.cpp
```diff
@@ -38,8 +39,13 @@ nlohmann::json Anchor::PrepRoomState() {
...
     payload["romhackName"] = Lighthouse::CurrentRomhackLabel();
-    payload["isRando"] = (bool)IS_RANDO;
-    payload["seed"] = (int32_t)(IS_RANDO ? RANDO_SEED : 0);
+    if (selectedFileNum != DEFAULT_FILE_NUM) {
+        payload["isRando"] = (bool)IS_RANDO;
+        payload["seed"] = (int32_t)(IS_RANDO ? RANDO_SEED : 0);
+    } else {
+        payload["isRando"] = roomState.isRando;
+        payload["seed"] = roomState.seed;
+    }
```
Malkierian, line 48: Now that I'm thinking of it, is this needed anymore since the OnGameLoad hook was changed to OnGameStart?

JeodC: Still needed; the hook change only moves when we request team state, it doesn't touch anything that publishes room state.

`PrepRoomState()` gets called from three places that don't go through that hook: `SendPacket_Handshake()` (we connect from file select, so no file num yet), the room settings widgets in `Menu.cpp`, and the romhack owner-refresh this PR adds, which isn't gated on a save being loaded. That last one would zero the room's seed just to fix the romhack label.

`CheckRandoRoomCompatibility()` is the only caller that early-outs on `!IsSaveLoaded()`, and it's the only place where the guard doesn't do anything.

PR #545, opened 2026-08-22
Add disable culling option

Pizzabucket:
## Summary

- Adds a "Disable Culling" graphics option.
- Disables viewport, model, and actor culling when the option is enabled.
- Preserves the required culling behavior during demo/playback sequences to prevent desynchronization.
- Includes a warning that disabling culling may affect performance.

## Credits

Parts of the demo/playback synchronization approach were adapted from BanjoRecomp.

## AI Use

AI was used in development, debugging, and build verification. I manually tested and published this PR.

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/10115899344.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/10115549473.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/10115517111.zip)

### DIFF
--- a/src/CC/ma/clanker.c
+++ b/src/CC/ma/clanker.c
@@ -66,24 +66,6 @@ enum maClankerState_e {
     MACLANKER_STATE_3_IDLE_RAISED
 };
 
-// LIGHTHOUSE_ALWAYS_LOAD_CLANKER_CC_V1
-// LIGHTHOUSE_ALWAYS_LOAD_CLANKER_CC_V2
-extern s32 getGameMode(void);
-
-static s32 lighthouse_clankerInDemoPlayback(void) {
-    switch (getGameMode()) {
-        case GAME_MODE_2_UNKNOWN:
-        case GAME_MODE_6_FILE_PLAYBACK:
-        case GAME_MODE_7_ATTRACT_DEMO:
-        case GAME_MODE_8_BOTTLES_BONUS:
-        case GAME_MODE_9_BANJO_AND_KAZOOIE:
-        case GAME_MODE_A_SNS_PICTURE:
-            return TRUE;
-        default:
-            return FALSE;
-    }
-}
-
 BKCollisionTriangle *__maClanker_getClankerCollisionTris(f32 arg0[3], f32 arg1[3], f32 arg2[3], u32 arg3){ // [port] pointer-width args
     BKCollisionTriangle *collision_tris;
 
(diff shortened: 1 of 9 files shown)


Pizzabucket:
Sorry for the last update. I broke the fading effect and the fixed hud models on the last one.

JeodC:
I'm going to be thorough in the review here, so don't take offense to anything. It's all for your information so you can improve. Thanks again for contributing some changes that could only be tested on unique hardware.

First, the PR description doesn't match the diff. "Fixes the scaling of 3D HUD objects at different aspect ratios" is claimed, but there's no scaling changes anywhere in the nine files. Either it was mistakenly dropped, or removed intentionally to be a separate, smaller pull request (which would be preferred anyway).

Second, you should cleanup your code before expecting a line-level review. Some examples:
- clanker.c:69-70: bare  bare `// LIGHTHOUSE_ALWAYS_LOAD_CLANKER_CC_V1` / `V2` markers.
- actor_cubepropsystem.c:402-445: `// ALL36 CUSTOM ACTOR TEST START/END` wrapping a bare `switch` on 36 unnamed raw marker IDs, described in-comment as "marker IDs that were stable in separate A/B tests." This is unmaintainable and untraceable; nobody can tell later why 421 is in the list and 422 isn't.
- Version suffixes baked into shipping identifiers: `cullV2`, `SORT one-sided geometry test V1`, `model distance gates V3.1`, `Single-Pass Demo Object No-Cull V2`, `// V3: generic sprite-actor renderers`
- modelRender.c: I spot at least four dead static functions, all will throw `-Wunused-function`. There are similar ones in viewport.c:56-95.
- There are six copies of the same game-mode switch, and four of them were added to decomp files which should remain as clean as possible. We already have a `IsDemoMode()` and the code also uses `func_802E4A08()` and `getGameMode() != GAME_MODE_4_PAUSED` as third and fourth demo predicates in the same feature.
- gccube.c:683-685: this is a misindented block whose behavior contradicts its own comment. The `if (port_shouldDisableCullingCubeRange())` sits inside the `else` at the wrong indent level. The comment one line above says "Culling ON, or demo/playback: completely stock range behavior", and then the very next statement widens `width` to the full cube range anyway whenever the cvar is set and `IsDemoMode()` is `false`. Either the comment or the code is wrong.
- GraphicsPatches.cpp:103: `port_shouldDisableCullingCubeRange` bypasses the file's own cvar cache. Follow what `port_shouldDisableCulling()` returns, and use the macro defined 80 lines above.
- gccube.c:296 and more: this is decomp divergence, a one-line decomp condition was expanded into an identical 7-line block at many call sites, with none of the original lines preserved.
- gccube.c:606-617: the sprite prepass walks every cube in the level, every frame, on top of the widened draw traversal. The tooltip's "May cost performance" is understating it.

PR #573, opened 2026-08-29
Entrance Index and Warp Lists

Malkierian:
Adds all entrances to a table that includes map ID, warp ID and English name. Also added a warp list using all `MapEntry`s, grouped by level. Some information, like warpsByLevel, are created on startup rather than being static info, potentially redundant.

The process of creating this list also allowed creating/finishing warp enum sets, and correcting a few decomp issues (done thanks to the "current map entrance listing" that is included in the warps tab). This process also added a level name variable that listed more than what was in ShipUtils' `warpNameList`, so I combined the two and made `port_mapName` also use that via `map_getLevel`, which I also added a NULL guard to to return 0 for Unknown as indexed in `worldNameList`. This also allowed the removal of the "level name - 1" uses of the previous `warpNameList`.

Finally, this consolidated all warp stuff inside the Warps tab, including aMannus's new player position warp functionality. Once the settings framework is added, which would allow direct data structure storage and reading, custom warp functionality will also be added.

UX improvement suggestions welcome.

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9738595138.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9738456339.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9738350689.zip)


aMannus (approved):

aMannus:
Assume the code cleanup introduced the merge conflicts

PR #577, opened 2026-08-30
Add Always Update Actors option

Pizzabucket:
### Summary

* Adds an **"Always Update Actors"** graphics option.
* Keeps supported loaded actors updating beyond their usual update-distance gates when the option is enabled.
* Keeps some specific gameplay checks unchanged and disables this option during a demo so they don't desync.
* Adds object-specific changes so they don't break when far away.
* Includes a warning that always updating more actors may affect performance.

### AI Use

AI was used in development, debugging, and build verification. I manually tested and published this PR.

### Build Artifacts
  - [Lighthouse-windows.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9774882744.zip)
  - [Lighthouse-linux.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9774548853.zip)
  - [Lighthouse-mac.zip](https://nightly.link/HarbourMasters/Lighthouse/actions/artifacts/9774363319.zip)

### DIFF
--- a/src/CCW/ch/snarebear.c
+++ b/src/CCW/ch/snarebear.c
@@ -2,6 +2,7 @@
 #include <ultra64.h>
 #include "functions.h"
 #include "variables.h"
+#include "port/Enhancements/Events/Hooks/Events.h"
 
 
 void chSnarebear_update(Actor *this);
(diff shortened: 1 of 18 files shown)


JeodC:
As I recall, ProxySaw had some problems with this in the recomp. Locking this behind a toggle is a start, but it would still be good to narrow down what breaks and where so support channels are easier to respond to for common issues.

Pizzabucket:
The only bug I spotted was Conga twitching with it on, so I just reverted him back to vanilla.

